PR #23, opened 2024-01-05
Refactor to use its own Datadog client

Tang8330:
## Changes

It was previously using Transfer's Datadog client, which caused a panic since the context keys are no longer the "same" since they're no longer string-based.

We are exclusively using our own Datadog client now.

### DIFF
--- a/lib/mtr/mtr.go
+++ b/lib/mtr/mtr.go
@@ -24,13 +24,12 @@ func InjectDatadogIntoCtx(ctx context.Context, namespace string, tags []string, 
 	if err != nil {
 		logger.FromContext(ctx).WithError(err).Fatal("failed to create datadog client")
 	}
-	
+
 	datadogClient.Tags = tags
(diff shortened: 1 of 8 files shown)


lib/mtr/mtr.go
```diff
@@ -0,0 +1,49 @@
...
+	datadogClient, err := statsd.New(address)
+	if err != nil {
+		logger.FromContext(ctx).WithError(err).Fatal("failed to create datadog client")
+	}
+	
+	datadogClient.Tags = tags
+	datadogClient.Namespace = stringutil.Override(DefaultNamespace, namespace)
+	return context.WithValue(ctx, constants.MtrKey, &statsClient{
+		client: datadogClient,
+		// Don't sample yet.
```
nathan-artie, line 32: Is this comment correct?

Tang8330: Nope, I'll remove. Thanks for the flag

nathan-artie (approved):

PR #28, opened 2024-01-24
[Config] Standing up the PostgreSQL config

Tang8330:
## Changes

1. Changing all the ints that cannot be negative to `uint`
2. Adding a PostgreSQL config as a step for us to migrate away from `custom-snapshots`

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -8,4 +8,7 @@ const (
 	MtrKey    contextKey = "__mtr"
 )
 
-const DefaultLimit = 5_000
+const (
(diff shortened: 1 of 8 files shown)


constants/constants.go
```diff
@@ -5,6 +5,7 @@ type contextKey string
 const (
 	ConfigKey contextKey = "__cfg"
 	KafkaKey  contextKey = "__kafka"
-	LoggerKey contextKey = "__logger"
```
nathan-artie, line 8: Nice!

nathan-artie (approved):

PR #37, opened 2024-01-24
Add function to create a Datadog client

nathan-artie:


Tang8330 (approved):

lib/mtr/mtr.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/artie-labs/reader/lib/logger"
 )
 
-func InjectDatadogIntoCtx(ctx context.Context, namespace string, tags []string, samplingRate float64) context.Context {
+func New(namespace string, tags []string, samplingRate float64) Client {
```
Tang8330, line 16: Does it need to be public? `newClient` maybe?

PR #43, opened 2024-01-26
Add postgres table iterator

nathan-artie:
Other than changing the package this is identical to https://github.com/artie-labs/custom-snapshots/blob/master/process/table.go


lib/kafkalib/writer.go
```diff
@@ -34,10 +34,6 @@ func NewBatchWriter(ctx context.Context, cfg config.Kafka) (BatchWriter, error)
 	return BatchWriter{w, ctx, cfg}, nil
 }
 
-func (w *BatchWriter) MaxMessageSize() uint64 {
```
nathan-artie, line 37: Don't need anymore.

Tang8330 (approved):

PR #47, opened 2024-01-26
Pull config and kafka writer out of context

nathan-artie:
Also moves postgres.Run() to sources/postgres

### DIFF
--- a/lib/kafkalib/writer.go
+++ b/lib/kafkalib/writer.go
@@ -63,13 +63,13 @@ func buildKafkaMessages(cfg *config.Kafka, msgs []RawMessage) ([]kafka.Message, 
 	return result, nil
 }
 
-func (k *BatchWriter) Write(rawMsgs []RawMessage) error {
-	msgs, err := buildKafkaMessages(&k.cfg, rawMsgs)
+func (w *BatchWriter) Write(rawMsgs []RawMessage) error {
+	msgs, err := buildKafkaMessages(&w.cfg, rawMsgs)
 	if err != nil {
 		return fmt.Errorf("failed to build to kafka messages: %w", err)
 	}
 
-	chunkSize := k.cfg.GetPublishSize()
+	chunkSize := w.cfg.GetPublishSize()
 
 	b := NewBatch(msgs, chunkSize)
 	if batchErr := b.IsValid(); batchErr != nil {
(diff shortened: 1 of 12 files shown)


sources/dynamodb/dynamodb.go
```diff
@@ -23,13 +24,15 @@ type Store struct {
...
-	tableName string
-	streamArn string
-	batchSize uint
-	streams   *dynamodbstreams.DynamoDBStreams
-	storage   *offsets.OffsetStorage
-	shardChan chan *dynamodbstreams.Shard
+	tableName   string
+	streamArn   string
+	batchSize   uint
+	topicPrefix string
```
nathan-artie, line 32: Added `topicPrefix`.

lib/kafkalib/writer.go
```diff
@@ -26,12 +26,8 @@ type BatchWriter struct {
...
-	w, err := NewWriter(ctx, cfg)
-	if err != nil {
-		return BatchWriter{}, err
-	}
-	return BatchWriter{w, ctx, cfg}, nil
+func NewBatchWriter(ctx context.Context, cfg config.Kafka, writer *kafka.Writer) BatchWriter {
+	return BatchWriter{writer, ctx, cfg}
 }
 
 func (w *BatchWriter) reload() error {
```
Tang8330, line 33: Unrelated to this PR - but can we be consistent about receiver names? Some using `w`, `k`, etc for `*BatchWriter`

main.go
```diff
@@ -32,25 +60,23 @@ func main() {
...
 
-	var statsD *mtr.Client
-	if cfg.Metrics != nil {
-		slog.Info("Injecting datadog")
-		_statsD, err := mtr.New(cfg.Metrics.Namespace, cfg.Metrics.Tags, 0.5)
-		if err != nil {
-			logger.Fatal("Failed to create datadog client", slog.Any("err", err))
-		}
-		statsD = &_statsD
+	kafka, err := setUpKafka(ctx, cfg.Kafka)
```
Tang8330, line 70: `_kafka`? so it doesn't collide with the import

sources/dynamodb/dynamodb.go
Tang8330: Now that we're injecting a bunch of variables inside of *Store, can we add a Validate function? So we ensure things like topicPrefix can never be empty str

Tang8330 (approved):
lgtm % c

lib/kafkalib/writer.go
nathan-artie: Good catch!

sources/dynamodb/dynamodb.go
nathan-artie: 👍 I'll add one in a new PR.

PR #56, opened 2024-01-27
Logging improvements

nathan-artie:


lib/logger/logger.go
```diff
@@ -32,6 +32,11 @@ func NewLogger(settings *config.Settings) (*slog.Logger, bool) {
 }
 
 func Fatal(msg string, args ...interface{}) {
+	slog.Error(msg, args...)
+	os.Exit(1)
+}
+
+func Panic(msg string, args ...interface{}) {
```
nathan-artie, line 39: Changing `Fatal` to `Panic` so that we mimic 
- https://pkg.go.dev/log#Panic
- https://pkg.go.dev/log#Fatal

PR #58, opened 2024-01-27
Cleanup postgres table iterator

nathan-artie:
- Rename `limit` to `batchSize`
- Exit early when we get back an empty array of rows
- Guard against calling `Next()` on an iterator that's already completed
- Return a closed iterator when scanning a table without any rows

### DIFF
--- a/lib/postgres/iterator.go
+++ b/lib/postgres/iterator.go
@@ -27,16 +27,16 @@ type TableIterator struct {
 	done          bool
 }
 
-func LoadTable(db *sql.DB, table *config.PostgreSQLTable, statsD *mtr.Client, maxRowSize uint64) (TableIterator, error) {
+func LoadTable(db *sql.DB, table *config.PostgreSQLTable, statsD *mtr.Client, maxRowSize uint64) (*TableIterator, error) {
 	slog.Info("Loading configuration for table", slog.String("table", table.Name))
 
 	postgresTable := NewTable(table)
 	if err := postgresTable.RetrieveColumns(db); err != nil {
 		if NoRowsError(err) {
 			slog.Info("Table does not contain any rows, skipping...", slog.String("table", table.Name))
-			return TableIterator{done: true}, nil
+			return nil, nil
 		} else {
-			return TableIterator{done: true}, fmt.Errorf("failed to validate postgres: %w", err)
+			return nil, fmt.Errorf("failed to validate postgres: %w", err)
 		}
 	}
 
@@ -48,7 +48,7 @@ func LoadTable(db *sql.DB, table *config.PostgreSQLTable, statsD *mtr.Client, ma
 		slog.Any("batchSize", table.GetLimit()),
 	)
 
-	return TableIterator{
+	return &TableIterator{
 		db:            db,
 		batchSize:     table.GetLimit(),
 		statsD:        statsD,
@@ -59,7 +59,7 @@ func LoadTable(db *sql.DB, table *config.PostgreSQLTable, statsD *mtr.Client, ma
 }
 
 func (i *TableIterator) HasNext() bool {
-	return !i.done
+	return i != nil && !i.done
 }
 
 func (i *TableIterator) recordMetrics(start time.Time) {
@@ -72,8 +72,8 @@ func (i *TableIterator) recordMetrics(start time.Time) {
 }
 
 func (i *TableIterator) Next() ([]lib.RawMessage, error) {
-	if i.done {
-		return nil, fmt.Errorf("cannot call Next() on a closed iterator")
+	if !i.HasNext() {
+		return nil, nil
 	}
 
 	rows, err := i.postgresTable.StartScanning(i.db,
@@ -87,8 +87,7 @@ func (i *TableIterator) Next() ([]lib.RawMessage, error) {
 	}
 
 	i.firstRow = false
-	// If the number of rows returned is less than the batch size, we've reached the end of the table
-	// TODO: Can we just exit in this case?
+	// The reason why lastRow exists is because in the past, we had queries only return partial results but it wasn't fully done
 	i.lastRow = i.batchSize > uint(len(rows))
 
 	var result []lib.RawMessage
--- a/lib/postgres/iterator.go
+++ b/lib/postgres/iterator.go
@@ -18,7 +18,7 @@ const DefaultErrorRetries = 10
 
 type TableIterator struct {
 	db            *sql.DB
-	limit         uint
+	batchSize     uint
 	statsD        *mtr.Client
 	maxRowSize    uint64
 	postgresTable *Table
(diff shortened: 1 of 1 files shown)


lib/postgres/iterator.go
```diff
@@ -28,29 +28,29 @@ type TableIterator struct {
 }
 
 func LoadTable(db *sql.DB, table *config.PostgreSQLTable, statsD *mtr.Client, maxRowSize uint64) (TableIterator, error) {
-	slog.Info("Loading configuration for table", slog.String("table", table.Name), slog.Any("limitSize", table.GetLimit()))
```
nathan-artie, line 31: Moved this to https://github.com/artie-labs/reader/pull/58/files#diff-84598fc3892e3d23cec5439e2a35ab3f99d5816d2302924ff22c4fdcf6162896R48

lib/postgres/iterator.go
```diff
@@ -28,29 +28,29 @@ type TableIterator struct {
...
+	slog.Info("Loading configuration for table", slog.String("table", table.Name))
 
 	postgresTable := NewTable(table)
-	err := postgresTable.RetrieveColumns(db)
-	if err != nil {
+	if err := postgresTable.RetrieveColumns(db); err != nil {
 		if NoRowsError(err) {
 			slog.Info("Table does not contain any rows, skipping...", slog.String("table", table.Name))
-			return TableIterator{}, nil
+			return TableIterator{done: true}, nil
```
nathan-artie, line 37: There was a little bug here in that we were returning an iterator that was not done so `HasNext()` would return `true` but calling `Next()` would return a NPE.

lib/postgres/iterator.go
```diff
@@ -62,23 +62,35 @@ func (i *TableIterator) HasNext() bool {
 	return !i.done
 }
 
-func (i *TableIterator) statsDTags() map[string]string {
-	return map[string]string{
-		"table":  strings.ReplaceAll(i.postgresTable.Name, `"`, ``),
-		"schema": i.postgresTable.Schema,
+func (i *TableIterator) recordMetrics(start time.Time) {
```
nathan-artie, line 65: Just moved the statsD code into this method.

lib/postgres/iterator.go
```diff
@@ -62,23 +62,35 @@ func (i *TableIterator) HasNext() bool {
...
+		(*i.statsD).Timing("scanned_and_parsed", time.Since(start), map[string]string{
+			"table":  strings.ReplaceAll(i.postgresTable.Name, `"`, ``),
+			"schema": i.postgresTable.Schema,
+		})
 	}
 }
 
 func (i *TableIterator) Next() ([]lib.RawMessage, error) {
-	var rows []map[string]interface{}
+	if i.done {
```
nathan-artie, line 75: Added guard against calling `Next` on an already completed iterator.

lib/postgres/iterator.go
```diff
@@ -62,23 +62,35 @@ func (i *TableIterator) HasNext() bool {
...
+		return nil, fmt.Errorf("cannot call Next() on a closed iterator")
+	}
+
 	rows, err := i.postgresTable.StartScanning(i.db,
-		NewScanningArgs(i.postgresTable.PrimaryKeys, i.limit, DefaultErrorRetries, i.firstRow, i.lastRow))
+		NewScanningArgs(i.postgresTable.PrimaryKeys, i.batchSize, DefaultErrorRetries, i.firstRow, i.lastRow))
 	if err != nil {
 		return nil, fmt.Errorf("failed to scan postgres: %w", err)
+	} else if len(rows) == 0 {
```
nathan-artie, line 83: Moved this from https://github.com/artie-labs/reader/pull/58/files#diff-84598fc3892e3d23cec5439e2a35ab3f99d5816d2302924ff22c4fdcf6162896L119

`i.limit` will always be greater than 0 because in the case that the table config has `limit` set to 0 we default it to 5,000. Therefore in the case that rows is empty `i.limit > uint(len(rows))` evaluates to true and we then check that rows is empty a second time. Instead of doing this we can just move the check here and exit early in the case that rows is empty.

lib/postgres/iterator.go
```diff
@@ -62,23 +62,35 @@ func (i *TableIterator) HasNext() bool {
...
+		slog.Info("Finished scanning", slog.String("table", i.postgresTable.Name))
+		i.done = true
+		return make([]lib.RawMessage, 0), nil
 	}
 
 	i.firstRow = false
-	var msgs []lib.RawMessage
+	// If the number of rows returned is less than the batch size, we've reached the end of the table
+	i.lastRow = i.batchSize > uint(len(rows))
```
nathan-artie, line 91: Moved this here from https://github.com/artie-labs/reader/pull/58/files#diff-84598fc3892e3d23cec5439e2a35ab3f99d5816d2302924ff22c4fdcf6162896L119

Since we will have returned if `rows` is empty the logic:
```
if i.limit > uint(len(rows)) {
  i.lastRow = true
} else {
  i.lastRow = false
}
```
Simplifies down to this.

Tang8330 (approved):

lib/postgres/iterator.go
```diff
@@ -28,29 +28,29 @@ type TableIterator struct {
 }
 
 func LoadTable(db *sql.DB, table *config.PostgreSQLTable, statsD *mtr.Client, maxRowSize uint64) (TableIterator, error) {
-	slog.Info("Loading configuration for table", slog.String("table", table.Name), slog.Any("limitSize", table.GetLimit()))
+	slog.Info("Loading configuration for table", slog.String("table", table.Name))
```
Tang8330, line 31: Can we return a ptr of `TableIterator`?

That way, you can just return `return nil, nil` vs `TableIteratoe{done: true}, nil`

Tang8330: Then:

```go
func (i *TableIterator) HasNext() bool {
    return i != nil && !i.done
}
```

lib/postgres/iterator.go
Tang8330: Call `.HasNext()` and if false then just return? So we can encapsulate this logic in 1 f(x)

lib/postgres/iterator.go
Tang8330: btw the reason why lastRow exists is because in the past, we had queries only return partial results but it wasn't fully done.

nathan-artie: As a follow up, I feel like when we've read the last row we can also set `done` to true since the next scan will always return an empty array of rows.

lib/postgres/iterator.go
nathan-artie: I like that.

lib/postgres/iterator.go
nathan-artie: Ah, that seems odd..., I'll add that as a comment.

PR #62, opened 2024-01-29
Use a copy of the primary keys when scanning a table

nathan-artie:
Scanning a table mutates the primary keys so we shouldn't be directly passing in the table's primary keys.

### DIFF
--- a/lib/postgres/primary_key/primary_keys.go
+++ b/lib/postgres/primary_key/primary_keys.go
@@ -55,7 +55,7 @@ func (k *Keys) Length() int {
 	return len(k.keys)
 }
 
-func (k *Keys) Copy() *Keys {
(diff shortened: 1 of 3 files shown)


lib/postgres/primary_key/primary_keys.go
```diff
@@ -55,6 +55,17 @@ func (k *Keys) Length() int {
 	return len(k.keys)
 }
 
+func (k *Keys) Copy() *Keys {
```
Tang8330, line 58: Perhaps `Clone()` would be slightly more idiomatic?

Tang8330 (approved):

lib/postgres/primary_key/primary_keys.go
nathan-artie: I like it.

PR #71, opened 2024-02-01
Clean up Kafka batch writer

nathan-artie:
- Pass context as a parameter
- Add a generic batch iterator and use it instead of `NewBatch`

### DIFF
--- a/lib/iterator/iterator.go
+++ b/lib/iterator/iterator.go
@@ -1,0 +1,41 @@
+package iterator
+
+type ArrayIterator[T any] interface {
+	HasNext() bool
+	Next() []T
+}
+
+func Collect[T any](iter ArrayIterator[T]) []T {
+	result := make([]T, 0)
+	for iter.HasNext() {
+		result = append(result, iter.Next()...)
+	}
+	return result
+}
+
+type batchIterator[T any] struct {
+	items []T
+	index int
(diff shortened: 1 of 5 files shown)


lib/iterator/iterator.go
```diff
@@ -0,0 +1,29 @@
+package iterator
+
+type batchIterator[T any] struct {
+	items []T
+	index int
+	n     int
```
Tang8330, line 6: What's n?

nathan-artie: size lol, I can rename.

lib/iterator/iterator.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+}
+
+func (i *batchIterator[T]) HasNext() bool {
+	return i.index < len(i.items)
+}
+
+func (i *batchIterator[T]) Next() []T {
+	if !i.HasNext() {
+		return make([]T, 0)
```
Tang8330, line 23: return nil?

lib/iterator/iterator.go
Tang8330: Seems like it's the batch size, maybe call it size or step instead?

Tang8330 (approved):
lgtm % c

PR #72, opened 2024-02-01
Use batch writer for dynamodb

nathan-artie:


lib/kafkalib/writer.go
```diff
@@ -75,11 +81,16 @@ func (w *BatchWriter) WriteMessages(ctx context.Context, msgs []kafka.Message) e
 
 	iter := iterator.NewBatchIterator(msgs, int(chunkSize))
 	for iter.HasNext() {
+		tags := map[string]string{
```
nathan-artie, line 84: Compare to https://github.com/artie-labs/reader/pull/72/files#diff-6f4928a6ebaf5212309eb572448fa23c7f13d15f9f3045f11f27b0043ec40861L67

lib/kafkalib/batch.go
```diff
@@ -1,103 +0,0 @@
...
-	"log/slog"
-	"time"
-
-	"github.com/artie-labs/transfer/lib/jitter"
-	"github.com/segmentio/kafka-go"
-
-	"github.com/artie-labs/reader/lib/mtr"
-)
-
-const (
```
nathan-artie, line 15: @Tang8330 by using the same code to write batches for both dyamodb and postgres we'll end up using the same constants, do we want to change them or just have specific constants for both?

For dynamodb:
- MaxRetries: 5
- RetryDelayMs: 250
- MaxMilliSeconds: 3500

For postgres:
- MaxRetries: 10
- RetryDelayMs: 300
- MaxMilliSeconds: 5000

sources/dynamodb/dynamodb.go
```diff
@@ -53,10 +50,8 @@ func Load(cfg config.Settings, statsD *mtr.Client, writer *kafka.Writer) (*Store
 	store := &Store{
 		tableName:   cfg.DynamoDB.TableName,
 		streamArn:   cfg.DynamoDB.StreamArn,
-		batchSize:   cfg.Kafka.GetPublishSize(),
```
nathan-artie, line 56: Don't need this since we read it in the writer: https://github.com/artie-labs/reader/pull/72/files#diff-8df5b95eaaa5d3b3344ee3ae5d980481037edbe98ab3da1dd6d5660ad5d80332R73

Tang8330 (approved):

PR #75, opened 2024-02-01
Fix error message and change logged attempts

nathan-artie:


lib/kafkalib/writer.go
```diff
@@ -97,7 +97,7 @@ func (w *BatchWriter) WriteMessages(ctx context.Context, msgs []kafka.Message) e
 				sleepMs := lib.JitterMs(baseJitterMs, maxJitterMs, attempts-1)
 				slog.Info("Failed to publish to kafka",
 					slog.Any("err", kafkaErr),
-					slog.Int("attempts", attempts-1),
+					slog.Int("attempts", attempts),
```
Tang8330, line 100: Intentional?

Tang8330 (approved):

lib/kafkalib/writer.go
nathan-artie: Yep, we've already made one attempt so outputting 1 seems more correct than 0.

PR #82, opened 2024-02-06
MongoDB snapshots

Tang8330:
## Changes

Implementing MongoDB snapshots to have feature parity with Debezium. 

## Checks

* [x] Test every data type
* [x] Include example
* [x] Add tests

## Data Types check 

[Comparison Gist](https://gist.github.com/Tang8330/d279bf33add5afd1245f4314edb6519c)

- [x] ObjectID
- [x] String
- [x] Int32
- [ ] Int64 (slight differences between SDKs)
- [x] Double
- [x] Decimal
- [x] SubDocument
- [x] Array
- [x] Timestamp
- [ ] Datetime (slight differences between the SDKs)
- [x] True
- [x] False
- [x] Null

PR #84, opened 2024-02-08
Refactoring Reader prior to MongoDB 

Tang8330:
MongoDB snapshots is done: #82, to make it easier to review, I split out non-MongoDB specific changes out so it's easier to review.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -68,8 +68,8 @@ const (
 
 type Settings struct {
 	Source     Source      `yaml:"source"`
-	PostgreSQL *PostgreSQL `yaml:"postgresql"`
-	DynamoDB   *DynamoDB   `yaml:"dynamodb"`
+	PostgreSQL *PostgreSQL `yaml:"postgresql,omitempty"`
+	DynamoDB   *DynamoDB   `yaml:"dynamodb,omitempty"`
 
 	Reporting *Reporting `yaml:"reporting"`
 	Metrics   *Metrics   `yaml:"metrics"`
--- a/lib/kafkalib/message.go
+++ b/lib/kafkalib/message.go
@@ -3,11 +3,10 @@ package kafkalib
 import (
 	"encoding/json"
 
-	"github.com/artie-labs/transfer/lib/cdc/util"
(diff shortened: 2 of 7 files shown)


config/config.go
```diff
@@ -68,8 +68,8 @@ const (
 
 type Settings struct {
 	Source     Source      `yaml:"source"`
-	PostgreSQL *PostgreSQL `yaml:"postgresql"`
-	DynamoDB   *DynamoDB   `yaml:"dynamodb"`
+	PostgreSQL *PostgreSQL `yaml:"postgresql,omitempty"`
```
Tang8330, line 71: Adding `omitempty` so that the generated diff if PostgreSQL or DynamoDB is empty, it'll be less.

lib/types.go
```diff
@@ -5,5 +5,17 @@ import "github.com/artie-labs/transfer/lib/cdc/util"
...
+
+func NewRawMessage(topicSuffix string, partitionKey map[string]interface{}, payload util.SchemaEventPayload) RawMessage {
+	return RawMessage{
+		TopicSuffix:  topicSuffix,
+		PartitionKey: partitionKey,
+		payload:      payload,
+	}
+}
+
+func (r RawMessage) GetPayload() interface{} {
```
Tang8330, line 19: This is an interface{} now because MongoDB's payload isn't `util.SchemaEventPayload`

lib/kafkalib/message.go
```diff
@@ -3,11 +3,10 @@ package kafkalib
 import (
 	"encoding/json"
 
-	"github.com/artie-labs/transfer/lib/cdc/util"
 	"github.com/segmentio/kafka-go"
 )
 
-func NewMessage(topic string, partitionKey map[string]interface{}, value util.SchemaEventPayload) (kafka.Message, error) {
+func NewMessage(topic string, partitionKey map[string]interface{}, value interface{}) (kafka.Message, error) {
```
Tang8330, line 9: Ditto - MongoDB doesn't use `util.SchemaEventPayload`

nathan-artie (approved):

PR #88, opened 2024-02-09
Postgres config improvements

nathan-artie:
- `port` is now defined as a `uint16` which aligns perfectly with the TCP/UDP port range of 0 to 65,535
- `userName` renamed to `username` (`userName` is still currently supported as a fallback)
- For table config `limit` renamed to `batchSize` (`limit` is still currently supported as a fallback)


Tang8330 (approved):
LGTM

PR #90, opened 2024-02-10
Pull Kafka writer out of DynamoDB `Source`

nathan-artie:
Instead just pass `writer` in the `Run` method.


Tang8330 (approved):
code lgtm, pipe some data through our test DDB and confirm that it works?

nathan-artie:
> code lgtm, pipe some data through our test DDB and confirm that it works?

Yep seems to work!

PR #97, opened 2024-02-13
Move `ColKindToDataType` to `columns.go`

nathan-artie:


lib/postgres/columns.go
```diff
@@ -53,3 +65,75 @@ func (t *Table) RetrieveColumns(db *sql.DB) error {
 
 	return t.FindStartAndEndPrimaryKeys(db)
 }
+
+func ColKindToDataType(colKind string, precision, scale *string, udtName *string) (debezium.DataType, *debezium.Opts) {
```
nathan-artie, line 69: No changes here.

lib/postgres/config.go
```diff
@@ -24,88 +22,8 @@ func NewPostgresConfig() *Config {
...
-			}
-		}
-	}
-
-	return debezium.InvalidDataType, nil
-}
-
-func (c *Config) AddColumn(colName, colKind string, precision, scale *string, udtName *string) {
-	dataType, opts := colKindToDebezium(colKind, precision, scale, udtName)
-	if dataType == debezium.InvalidDataType {
```
nathan-artie, line 101: Moved this logic to https://github.com/artie-labs/reader/pull/97/files#diff-35c5e80d88acd84f93f45c188bc0826f3bc869f1ddd0cc66134b98f78234a66cL34

lib/postgres/columns.go
```diff
@@ -31,7 +35,15 @@ func (t *Table) RetrieveColumns(db *sql.DB) error {
 			return err
 		}
 
-		t.Config.AddColumn(colName, colKind, numericPrecision, numericScale, udtName)
+		dataType, opts := ColKindToDataType(colKind, numericPrecision, numericScale, udtName)
```
Tang8330, line 38: make this a priv func? `ColKindToDataType`

Tang8330 (approved):
:shipit:

lib/postgres/columns.go
nathan-artie: Ooh I thought it was in a different package than some of our tests that use it but actually it isn't.

PR #100, opened 2024-02-13
Kill postgres table `Config` struct

nathan-artie:
Just store `Fields` directly on `Table`.


lib/postgres/config_test.go
```diff
@@ -5,6 +5,8 @@ import (
 
 	"github.com/artie-labs/transfer/lib/debezium"
 	"github.com/stretchr/testify/assert"
+
+	pgDebezium "github.com/artie-labs/reader/lib/postgres/debezium"
 )
 
 func TestPostgresConfig_Complete(t *testing.T) {
```
nathan-artie, line 12: This file is now very poorly named, and it's testing a lot, I'll try to decompose it in a later PR.

Tang8330 (approved):

PR #106, opened 2024-02-15
Add a stateless postgres describe table function

nathan-artie:
Splits out the stateless part of describing a postgres table + determining the column types from adding fields to our stateful `Table` struct.


Tang8330 (approved):
Love it

PR #107, opened 2024-02-15
Using Null metrics provider if none is available

Tang8330:
## Changes

1. Using `Null Metrics Provider` from `Transfer` if metrics was not passed in.
2. Cleaning up code to not rely on a pointer to an interface

### DIFF
--- a/lib/mtr/datadog.go
+++ b/lib/mtr/datadog.go
@@ -16,7 +16,7 @@ const (
 type Client interface {
 	Timing(name string, value time.Duration, tags map[string]string)
 	Incr(name string, tags map[string]string)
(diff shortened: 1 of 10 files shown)


lib/mtr/datadog.go
```diff
@@ -16,7 +16,7 @@ const (
 type Client interface {
 	Timing(name string, value time.Duration, tags map[string]string)
 	Incr(name string, tags map[string]string)
-	Gauge(name string, tags map[string]string, value float64)
+	Gauge(name string, value float64, tags map[string]string)
```
Tang8330, line 19: Doing this so we can conform to Transfer's Client interface.

nathan-artie (approved):
Nice!

PR #110, opened 2024-02-15
[postgres] Add functions to get the bounds of primary keys

nathan-artie:


lib/postgres/schema/schema.go
```diff
@@ -204,3 +205,47 @@ func GetPrimaryKeys(db *sql.DB, schema, table string) ([]string, error) {
 	}
 	return primaryKeys, nil
 }
+
+func GetPrimaryKeysLowerBounds(db *sql.DB, schema, table string, primaryKeys []string, castedPrimaryKeys []string) ([]interface{}, error) {
+	minValues := make([]interface{}, len(primaryKeys))
```
nathan-artie, line 210: @Tang8330 do we need to be making two slices here, seems like one would suffice?

nathan-artie: Ohh I see, one is a slice of pointers.

lib/postgres/schema/schema_test.go
```diff
@@ -126,3 +126,26 @@ func TestParseColumnDataType(t *testing.T) {
 		assert.Equal(t, testCase.expectedOpts, opts, testCase.name)
 	}
 }
+
+func TestSelectTableQuery(t *testing.T) {
```
nathan-artie, line 130: Just moved this.

lib/postgres/schema/schema.go
```diff
@@ -204,3 +204,68 @@ func GetPrimaryKeys(db *sql.DB, schema, table string) ([]string, error) {
...
+
+type selectTableQueryArgs struct {
+	Keys       []string
+	Schema     string
+	TableName  string
+	OrderBy    []string
+	Descending bool
+}
+
+func selectTableQuery(args selectTableQueryArgs) string {
```
nathan-artie, line 216: Just moved this function, no changes.

Tang8330 (approved):

PR #111, opened 2024-02-15
[postgres] Add function to get both upper and lower pk bounds

nathan-artie:


lib/postgres/table.go
```diff
@@ -46,17 +45,11 @@ func (t *Table) TopicSuffix() string {
 }
 
 func (t *Table) FindStartAndEndPrimaryKeys(db *sql.DB) error {
-	primaryKeys, err := schema.GetPrimaryKeys(db, t.Schema, t.Name)
+	keys, err := schema.GetPrimaryKeys(db, t.Schema, t.Name)
 	if err != nil {
 		return fmt.Errorf("failed to retrieve primary keys: %w", err)
 	}
 
-	for _, primaryKey := range primaryKeys {
```
nathan-artie, line 54: We now just call `Upsert` once per primary key.

Tang8330 (approved):

PR #113, opened 2024-02-15
[postgres] Kill extra query for table columns

nathan-artie:
I don't think we need to query for columns again since we already have them from when we described the table.

### DIFF
--- a/lib/postgres/columns.go
+++ b/lib/postgres/columns.go
@@ -9,7 +9,7 @@ import (
 	"github.com/jackc/pgx/v5"
 )
 
-func (t *Table) RetrieveColumns(db *sql.DB) error {
+func (t *Table) PopulateColumns(db *sql.DB) error {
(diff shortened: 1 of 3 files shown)


lib/postgres/columns.go
```diff
@@ -23,24 +23,10 @@ func (t *Table) RetrieveColumns(db *sql.DB) error {
...
-	if err != nil {
-		return err
-	}
 
-	for _, column := range columns {
 		// Add to original columns before mutation
-		t.OriginalColumns = append(t.OriginalColumns, column)
-		columnKind := t.Fields.GetDataType(column)
-		t.ColumnsCastedForScanning = append(t.ColumnsCastedForScanning, castColumn(column, columnKind))
+		t.OriginalColumns = append(t.OriginalColumns, col.Name)
```
nathan-artie, line 28: @Tang8330 Do we want to be setting these for columns that we're ignoring because we can't parse the data type?

Tang8330 (approved):
lgtm

PR #119, opened 2024-02-15
[postgres] Change `castColumn` to accept `Column` struct

nathan-artie:
- Change `castColumn` to accept `Column`


lib/postgres/queries/queries.go
```diff
@@ -33,7 +34,8 @@ type ScanTableQueryArgs struct {
 	Schema        string
 	TableName     string
 	PrimaryKeys   []string
-	ColumnsToScan []string
+	ColumnsToScan []schema.Column
+	ColumnCast    func(schema.Column) string
```
Tang8330, line 38: Why does this need to be an arg function? Can't we just call the function inline?

Tang8330 (approved):

lib/postgres/queries/queries.go
nathan-artie: Reverting this for now.

PR #124, opened 2024-02-15
[postgres] Move primary key casting to schema

nathan-artie:


lib/postgres/schema/schema.go
```diff
@@ -206,38 +206,43 @@ func GetPrimaryKeys(db *sql.DB, schema, table string) ([]string, error) {
 }
 
 type selectTableQueryArgs struct {
-	Keys       []string
+	Keys       []Column
 	Schema     string
 	TableName  string
-	OrderBy    []string
 	Descending bool
+	CastFunc   func(c Column) string
```
nathan-artie, line 213: I decided to pass a function here because I don't want the schema to know about `castColumn` since `castColumn` is specific to how we're currently querying for data and not general to all postgres queries.

lib/postgres/schema/schema.go
```diff
@@ -206,38 +206,43 @@ func GetPrimaryKeys(db *sql.DB, schema, table string) ([]string, error) {
 }
 
 type selectTableQueryArgs struct {
-	Keys       []string
+	Keys       []Column
 	Schema     string
 	TableName  string
-	OrderBy    []string
```
nathan-artie, line 212: `OrderBy` was always the same as `Keys` except that the values were uncasted.

Tang8330 (approved):

PR #132, opened 2024-02-17
[postgres] Add `ColumnToField` function

nathan-artie:


lib/postgres/debezium/fields.go
```diff
@@ -52,28 +52,31 @@ func (f *Fields) GetDataType(fieldName string) schema.DataType {
 	return dataType
 }
 
-func (f *Fields) AddField(colName string, dataType schema.DataType, opts *schema.Opts) {
-	res := ToDebeziumType(dataType)
+func ColumnToField(col schema.Column) debezium.Field {
```
nathan-artie, line 55: Will move this somewhere else later but putting it here for now so the diff is easier to see.

PR #137, opened 2024-02-17
[postgres] Move `KeysToValueList` calls to `scanTableQuery`

nathan-artie:
`KeysToValueList ` is postgres specific where as the rest of the primary key code is database agnostic and can be shared with mysql. Will move it to `lib/rdbms` after this.

### DIFF
--- a/lib/postgres/primary_key/primary_keys.go
+++ b/lib/postgres/primary_key/primary_keys.go
@@ -3,14 +3,6 @@ package primary_key
 import (
 	"fmt"
 	"strings"
-
-	"github.com/artie-labs/transfer/lib/cdc"
-	"github.com/artie-labs/transfer/lib/cdc/util"
-	"github.com/artie-labs/transfer/lib/debezium"
-	"github.com/artie-labs/transfer/lib/typing"
-
-	pgDebezium "github.com/artie-labs/reader/lib/postgres/debezium"
-	"github.com/artie-labs/reader/lib/postgres/schema"
 )
 
 type Keys struct {
@@ -113,38 +105,6 @@ func (k *Keys) Keys() []string {
 	return keysToReturn
 }
 
-// TODO: This function should just fold into the column escape function.
-func (k *Keys) KeysToValueList(columns []schema.Column, end bool) []string {
-	optionalSchema := getOptionalSchema(columns)
-
-	var valuesToReturn []string
-	for _, pk := range k.keys {
-		val := pk.StartingValue
-		if end {
-			val = pk.EndingValue
-		}
-
-		kindDetails := typing.ParseValue(typing.Settings{}, pk.Name, optionalSchema, val)
-		switch kindDetails.Kind {
-		case typing.String.Kind, typing.Struct.Kind, typing.ETime.Kind:
-			valuesToReturn = append(valuesToReturn, fmt.Sprintf(`'%s'`, val))
-		default:
-			valuesToReturn = append(valuesToReturn, val)
-		}
-	}
-
-	return valuesToReturn
-}
-
-func getOptionalSchema(columns []schema.Column) map[string]typing.KindDetails {
-	schemaEvtPayload := &util.SchemaEventPayload{
-		Schema: debezium.Schema{
-			FieldsObject: []debezium.FieldsObject{{
-				Fields:     pgDebezium.ColumnsToFields(columns),
-				Optional:   false,
-				FieldLabel: cdc.After,
-			}},
-		},
-	}
-	return schemaEvtPayload.GetOptionalSchema()
+func (k *Keys) KeysList() []Key {
+	return k.keys
 }
--- a/lib/postgres/scan.go
+++ b/lib/postgres/scan.go
@@ -68,11 +68,9 @@ type scanTableQueryArgs struct {
 	Columns     []schema.Column
 
 	// First where clause
-	FirstWhere   Comparison
-	StartingKeys []string
+	FirstWhere Comparison
 	// Second where clause
 	SecondWhere Comparison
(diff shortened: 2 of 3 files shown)


lib/postgres/scan.go
```diff
@@ -74,8 +74,7 @@ type scanTableQueryArgs struct {
 	SecondWhere Comparison
 	EndingKeys  []string
 
-	OrderBy []string
```
nathan-artie, line 77: `OrderBy` is the same as `PrimaryKeys` so just pass one in.

lib/postgres/scan.go
```diff
@@ -107,23 +107,17 @@ func (s *scanner) scan(errorAttempts int) ([]map[string]interface{}, error) {
...
 	}
 
-	startKeys := s.primaryKeys.KeysToValueList(s.table.Columns, false)
-	endKeys := s.primaryKeys.KeysToValueList(s.table.Columns, true)
-
 	query := scanTableQuery(scanTableQueryArgs{
 		Schema:      s.table.Schema,
 		TableName:   s.table.Name,
-		PrimaryKeys: s.table.PrimaryKeys.Keys(),
+		PrimaryKeys: s.primaryKeys,
```
nathan-artie, line 113: It should be safe to use `s.primaryKeys` for `s.table.PrimaryKeys.Keys()` as they same the same keys, they just differ in values. The values are updated as the scanner scans through the table.

lib/postgres/scan.go
```diff
@@ -64,18 +64,15 @@ func (c Comparison) SQLString() string {
 type scanTableQueryArgs struct {
 	Schema      string
 	TableName   string
-	PrimaryKeys []string
+	PrimaryKeys *primary_key.Keys
 	Columns     []schema.Column
 
 	// First where clause
-	FirstWhere   Comparison
-	StartingKeys []string
```
nathan-artie, line 72: Compute `StartingKeys` and `EndingKeys` inside `scanTableQuery`

lib/postgres/scan.go
```diff
@@ -107,23 +107,17 @@ func (s *scanner) scan(errorAttempts int) ([]map[string]interface{}, error) {
...
 		Columns:     s.table.Columns,
 
-		FirstWhere:   firstWhereClause,
-		StartingKeys: startKeys,
+		FirstWhere: firstWhereClause,
 
 		SecondWhere: secondWhereClause,
-		EndingKeys:  endKeys,
 
-		OrderBy: s.table.PrimaryKeys.Keys(),
```
nathan-artie, line 125: `OrderBy` is the same as `PrimaryKeys`, so no need to pass both in.

lib/postgres/primary_key/primary_keys.go
```diff
@@ -113,38 +105,6 @@ func (k *Keys) Keys() []string {
...
-			valuesToReturn = append(valuesToReturn, fmt.Sprintf(`'%s'`, val))
-		default:
-			valuesToReturn = append(valuesToReturn, val)
-		}
-	}
-
-	return valuesToReturn
-}
-
-func getOptionalSchema(columns []schema.Column) map[string]typing.KindDetails {
```
nathan-artie, line 139: Moved this https://github.com/artie-labs/reader/pull/137/files#diff-eb76bea08e483c31dcfdd81272705c460694b4cc6fb60a396854c96d676dd9adR105 and inlined it.

lib/postgres/primary_key/primary_keys.go
```diff
@@ -113,38 +105,6 @@ func (k *Keys) Keys() []string {
 	return keysToReturn
 }
 
-// TODO: This function should just fold into the column escape function.
-func (k *Keys) KeysToValueList(columns []schema.Column, end bool) []string {
```
nathan-artie, line 117: Moved this https://github.com/artie-labs/reader/pull/137/files#diff-eb76bea08e483c31dcfdd81272705c460694b4cc6fb60a396854c96d676dd9adR116

lib/postgres/scan.go
```diff
@@ -84,18 +86,52 @@ func scanTableQuery(args scanTableQueryArgs) string {
...
+			FieldsObject: []debezium.FieldsObject{{
+				Fields:     pgDebezium.ColumnsToFields(columns),
+				Optional:   false,
+				FieldLabel: cdc.After,
+			}},
+		},
+	}
+	optionalSchema := schemaEvtPayload.GetOptionalSchema()
+
+	var valuesToReturn []string
```
nathan-artie, line 116: The only change here is that we do `for _, pk := range k.KeysList() {` instead of `for _, pk := range k.keys {` because `keys` is private.

lib/postgres/scan.go
```diff
@@ -84,18 +86,52 @@ func scanTableQuery(args scanTableQueryArgs) string {
...
-		strings.Join(QuotedIdentifiers(args.PrimaryKeys), ","), args.SecondWhere.SQLString(), strings.Join(args.EndingKeys, ","),
-		strings.Join(QuotedIdentifiers(args.OrderBy), ","),
+		strings.Join(QuotedIdentifiers(args.PrimaryKeys.Keys()), ","), args.SecondWhere.SQLString(), strings.Join(endingValues, ","),
+		strings.Join(QuotedIdentifiers(args.PrimaryKeys.Keys()), ","),
 		args.Limit,
 	)
 }
 
+func keysToValueList(k *primary_key.Keys, columns []schema.Column, end bool) []string {
+	schemaEvtPayload := &util.SchemaEventPayload{
```
nathan-artie, line 105: Inlined from here to line 114.

Tang8330 (approved):

PR #144, opened 2024-02-17
[postgres] Inline call to `GetOptionalSchema`

nathan-artie:


lib/postgres/scan.go
```diff
@@ -109,16 +106,14 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
 }
 
 func shouldQuoteValue(col schema.Column, val string) bool {
-	schemaEvtPayload := &util.SchemaEventPayload{
```
nathan-artie, line 112: Compare to `https://github.com/artie-labs/transfer/blob/master/lib/cdc/util/optional_schema.go#L10` - functionality should be identical.

Tang8330 (approved):

PR #145, opened 2024-02-17
[postgres] Refactor out call to `ParseValue`

nathan-artie:
Copied `ParseValue` https://github.com/artie-labs/transfer/blob/33b404f72a33c6577403728a8fe39570aa0ef558/lib/typing/typing.go#L102 to a temporary package and then slowly refactored it out. The resulting logic should be identical to what we had before. Please review each commit individually to ensure no subtle errors were introduced.


lib/postgres/temp/parse_value.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"reflect"
+	"strings"
+
+	"github.com/artie-labs/transfer/lib/typing"
+	"github.com/artie-labs/transfer/lib/typing/decimal"
+	"github.com/artie-labs/transfer/lib/typing/ext"
+)
+
+func ParseValue(settings typing.Settings, key string, optionalSchema map[string]typing.KindDetails, val interface{}) typing.KindDetails {
```
nathan-artie, line 12: Copied this from https://github.com/artie-labs/transfer/blob/33b404f72a33c6577403728a8fe39570aa0ef558/lib/typing/typing.go#L102 - the only changes are adding `typing.` to the types.

lib/postgres/scan.go
```diff
@@ -105,18 +105,13 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
...
-	}
-
-	kindDetails := typing.ParseValue(typing.Settings{}, col.Name, optionalSchema, val)
-	switch kindDetails.Kind {
-	case typing.String.Kind, typing.Struct.Kind, typing.ETime.Kind:
+func shouldQuoteValue(col schema.Column) bool {
+	kindDetail := pgDebezium.ColumnToField(col).ToKindDetails()
+	switch kindDetail.Kind {
+	case typing.Invalid.Kind:
+		slog.Warn("Could not determine data type for column", slog.String("colName", col.Name))
```
Tang8330, line 112: Invalid should be false.

lib/postgres/scan.go
```diff
@@ -105,18 +105,13 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
...
+		return true
+	case typing.String.Kind, typing.Struct.Kind, typing.ETime.Kind, typing.EDecimal.Kind:
```
Tang8330, line 114: EDecimal?

lib/postgres/scan.go
nathan-artie: That would change the behavior of this function as we currently quote invalid kinds.
When a kind is invalid:
- `pgDebezium.ColumnToField(col).ToKindDetails()` returns an invalid kind
- `optionalSchema` is then empty
-  The call to `ParseValue` ignores `optionalSchema` and hits the `case string:` block
- The `case string:` block returns either `ETime`, `Struct`, or `String`, all of which are quoted

lib/postgres/scan.go
nathan-artie: Yep, currently if we have a `EDecimal` in `optionalSchema`: 
- when we call `ParseValue` we hit https://github.com/artie-labs/transfer/blob/33b404f72a33c6577403728a8fe39570aa0ef558/lib/typing/typing.go#L112
- we then call `ParseValue` again this time passing in `nil` for `optionalSchema`
- we then hit the `case string:` block in `ParseValue`
- The `case string`: block returns either `ETime`, `Struct`, or `String`, all of which are quoted

lib/postgres/scan.go
Tang8330: Why would it hit string?

lib/postgres/scan.go
Tang8330: It wouldn't hit this? https://github.com/artie-labs/transfer/blob/cdf82a6e5b2ff966aeb1e1468fd149dc27417c21/lib/typing/typing.go#L157-L161

lib/postgres/scan.go
nathan-artie: Because the `val` argument we're passing to `ParseValue` is always a string when we call it from this function.

lib/postgres/scan.go
nathan-artie: The `val` argument we're passing to `ParseValue` is always a string when we call it from this function.

nathan-artie: This happens because we we're always storing primary keys as strings.

lib/postgres/scan.go
nathan-artie: This happens because we we're always storing primary keys as strings.

Tang8330 (approved):
sgtm

PR #147, opened 2024-02-18
[postgres] Modify `shouldQuoteValue` to check types directly

nathan-artie:
Based on https://github.com/artie-labs/reader/pull/145 I determined that the `val` passed to `shouldQuoteValue` is extraneous. Instead of the complexity of:
- converting a `Column` to a `Field`
- converting the `Field` to a `KindDetails`
  - For the decimals the existing behavior https://github.com/artie-labs/transfer/blob/cdf82a6e5b2ff966aeb1e1468fd149dc27417c21/lib/debezium/schema.go#L92 checks that we can parse the scale and if not returns in invalid type, however since we quoted invalid types and also quote decimal types this doesn't actually matter.
- switching on `KindDetails`

we can just do a simple switch on the `Column`'s `DataType`. I already have tests for all of the Postgres `DataType`s so we can be sure that this change preserves the existing behavior.

### DIFF
--- a/lib/postgres/scan.go
+++ b/lib/postgres/scan.go
@@ -107,10 +107,6 @@ func shouldQuoteValue(col schema.Column) (bool, error) {
 	switch col.Type {
 	case schema.InvalidDataType:
 		return false, fmt.Errorf("invalid data type")
-	case schema.Money, schema.Numeric:
-		// TODO: Returning true for these to maintain parity with existing code, but double check this
-		slog.Warn("Could not determine data type for column", slog.String("colName", col.Name))
-		return true, nil
 	case schema.Float,
 		schema.Int16,
 		schema.Int32,
@@ -121,6 +117,8 @@ func shouldQuoteValue(col schema.Column) (bool, error) {
 		schema.Array:    // TODO: Double check this
 		return false, nil
 	case schema.VariableNumeric,
+		schema.Money,
+		schema.Numeric,
 		schema.TextThatRequiresEscaping,
 		schema.Text,
 		schema.HStore,
--- a/lib/postgres/scan.go
+++ b/lib/postgres/scan.go
@@ -10,10 +10,8 @@ import (
 
 	"github.com/artie-labs/transfer/lib/jitter"
 	"github.com/artie-labs/transfer/lib/ptr"
-	"github.com/artie-labs/transfer/lib/typing"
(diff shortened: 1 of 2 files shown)


lib/postgres/scan.go
```diff
@@ -105,16 +103,37 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
...
-	switch kindDetail.Kind {
-	case typing.Invalid.Kind:
-		slog.Warn("Could not determine data type for column", slog.String("colName", col.Name))
-		return true
-	case typing.String.Kind, typing.Struct.Kind, typing.ETime.Kind, typing.EDecimal.Kind:
-		return true
+func shouldQuoteValue(col schema.Column) (bool, error) {
+	switch col.Type {
+	case schema.InvalidDataType:
+		return false, fmt.Errorf("invalid data type")
```
Tang8330, line 109: Expected?

Tang8330: I prefer this actually, it'll flag unknown data types as opposed to failing silently.

nathan-artie: Yep, we should never be calling this function with an `InvalidDataType`.

lib/postgres/scan.go
```diff
@@ -105,16 +103,37 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
...
+	case schema.Float,
+		schema.Int16,
+		schema.Int32,
+		schema.Int64,
+		schema.Bit,
+		schema.Boolean,
+		schema.Interval, // TODO: Double check this
```
Tang8330, line 116: Are you planning to check prior to merge?

nathan-artie: Nope, I'll do that as a FF. Not quoting `Interval` or `Array` is the current behavior though I suspect it's wrong.

Tang8330: k

Tang8330 (approved):

lib/postgres/scan.go
nathan-artie: When we load the columns we throw an error if we cannot determine any of the types: https://github.com/artie-labs/reader/blob/ed33aac4f28361ed59b28f6e3f0426cf6aad3d0f/lib/postgres/schema/schema.go#L96

PR #156, opened 2024-02-18
[postgres] Add adapter for `DebeziumTransformer`

nathan-artie:
This will allow the `DebeziumTransformer ` to be source agnostic so that it can be reused for `MySQL`. For this PR I tried to make as few changes as possible so that it would be easier to review. In the next PR I will move `DebeziumTransformer` to `lib/transformer`.

### DIFF
--- a/sources/postgres/transformer/transformer.go
+++ b/sources/postgres/transformer/transformer.go
@@ -14,17 +14,34 @@ import (
 	"github.com/artie-labs/reader/lib/postgres"
 )
 
-type DebeziumTransformer struct {
-	statsD mtr.Client
-	table  *postgres.Table
-	iter   batchRowIterator
+type Adapter interface {
+	TableName() string
+	TopicSuffix() string
+	PartitionKey(row map[string]interface{}) map[string]interface{}
+	Fields() []debezium.Field
+	MetricTags() map[string]string
+	ConvertRowToDebezium(row map[string]interface{}) (map[string]interface{}, error)
 }
 
-func NewDebeziumTransformer(table *postgres.Table, iter batchRowIterator, statsD mtr.Client) *DebeziumTransformer {
+type postgresAdapter struct {
+	table postgres.Table
+}
+
+func NewPostgresAdapter(table postgres.Table) Adapter {
+	return postgresAdapter{table: table}
+}
+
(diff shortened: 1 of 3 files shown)


sources/postgres/transformer/transformer.go
```diff
@@ -64,26 +84,26 @@ func (d *DebeziumTransformer) Next() ([]lib.RawMessage, error) {
...
 
-		result = append(result, lib.NewRawMessage(d.topicSuffix(), d.partitionKey(row), payload))
+		result = append(result, lib.NewRawMessage(d.adapter.TopicSuffix(), d.adapter.PartitionKey(row), payload))
 		d.recordMetrics(start)
 	}
 	return result, nil
 }
 
-func (d *DebeziumTransformer) topicSuffix() string {
+func (d postgresAdapter) TopicSuffix() string {
```
nathan-artie, line 93: I will rename `d` to `p` in a follow up PR.

sources/postgres/transformer/transformer.go
```diff
@@ -14,17 +14,34 @@ import (
 	"github.com/artie-labs/reader/lib/postgres"
 )
 
+type Adapter interface {
+	TableName() string
+	TopicSuffix() string
+	PartitionKey(row map[string]interface{}) map[string]interface{}
+	Fields() []debezium.Field
+	MetricTags() map[string]string
```
Tang8330, line 22: If it's easier, we can remove `metricTags` since we're not using them anyways.

We are using the ones from Kafka batch writer and when we start to run snapshots inline, we'll use the metrics provided by Transfer's `ProcessMessage(...)`

Tang8330: Which I assume is only being used by `recordMetrics`

Tang8330 (approved):

sources/postgres/transformer/transformer.go
nathan-artie: Ooh nice, will do!

nathan-artie: So we can remove `d.statsD.Timing("scanned_and_parsed", time.Since(start), d.adapter.MetricTags())` entirely or just the tags part?

PR #170, opened 2024-02-20
Allow primary key values to be any type

nathan-artie:
For MySQL we will just use the values returned from the database rather than coercing them to strings like we do for PostgreSQL.


lib/postgres/scan.go
```diff
@@ -155,10 +155,13 @@ func keysToValueList(k *primary_key.Keys, columns []schema.Column, end bool) ([]
 			return nil, err
 		}
 
+		// TODO: look into storing primary key values as their raw types instead of converting them to strings
+		strVal := val.(string)
```
Tang8330, line 159: type coercions like this can cause panics

better to do fmt.Sprint(val)

Tang8330 (approved):
lgtm % c

PR #181, opened 2024-02-21
[mysql] Fix primary keys bug; build scanner query

nathan-artie:


lib/mysql/schema/schema.go
```diff
@@ -267,7 +267,7 @@ func getPrimaryKeyValues(db *sql.DB, table string, primaryKeys []Column, descend
 	}
 	defer stmt.Close()
 
-	if err := stmt.QueryRow(query, 1).Scan(resultPtrs...); err != nil {
+	if err := stmt.QueryRow().Scan(resultPtrs...); err != nil {
```
nathan-artie, line 270: Don't need the `1` anymore and `query` is already being passed in online 264.

Tang8330 (approved):

PR #182, opened 2024-02-21
[PostgreSQL] Don't escape money.

Tang8330:
Money doesn't need to be escaped, compared our generated message against Debezium and looks correct.

### DIFF
--- a/lib/postgres/cast.go
+++ b/lib/postgres/cast.go
@@ -14,7 +14,7 @@ func castColumn(col schema.Column) string {
 	switch col.Type {
 	case schema.InvalidDataType:
 		return colName
-	case schema.Money, schema.TextThatRequiresEscaping:
+	case schema.TextThatRequiresEscaping:
 		return fmt.Sprintf("%s::text", colName)
 	case schema.Time, schema.Interval:
 		// We want to extract(epoch from interval) will emit this in ms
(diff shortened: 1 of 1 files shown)


lib/postgres/cast.go
```diff
@@ -30,8 +30,11 @@ func castColumn(col schema.Column) string {
 		return fmt.Sprintf(`cast(extract(epoch from %s)*%d as bigint) as "%s"`, colName, multiplier, col.Name)
 	case schema.Array:
 		return fmt.Sprintf(`ARRAY_TO_JSON(%s)::TEXT as "%s"`, colName, col.Name)
-	case schema.Int16, schema.Int32, schema.Int64, schema.UUID, schema.UserDefinedText,
-		schema.VariableNumeric, schema.Numeric, schema.Text, schema.Boolean, schema.Date, schema.Timestamp, schema.HStore, schema.JSON, schema.Bit:
+	case schema.Int16, schema.Int32, schema.Int64, schema.UUID,
```
Tang8330, line 33: Just added Money, and then organized the data types a bit based on types similarity

lib/postgres/scan_test.go
```diff
@@ -100,9 +100,9 @@ func TestScanTableQuery(t *testing.T) {
 			{Name: "c", Type: schema.Int64},
 			{Name: "e", Type: schema.Text},
 			{Name: "f", Type: schema.Int64},
-			{Name: "g", Type: schema.Money}, // money will be cast
+			{Name: "g", Type: schema.Money},
```
nathan-artie, line 103: The intention for having `schema.Money` here was to make sure we were calling cast, how about swapping `schema.Money` for `schema.TextThatRequiresEscaping`?

Tang8330: SGTM

PR #185, opened 2024-02-22
[postgres] Reorder scanner methods

nathan-artie:
Also added a test.


lib/postgres/scan.go
```diff
@@ -52,119 +52,27 @@ func (t *Table) NewScanner(db *sql.DB, batchSize uint, errorRetries int) (scanne
 	}, nil
 }
 
-type scanTableQueryArgs struct {
```
nathan-artie, line 55: No changes to this except moving things around.

PR #191, opened 2024-02-23
[rdbms] Change primary_key `Upsert` to `UpdateStartingValue`

nathan-artie:
Now that we're initializing the `Keys` struct with all the keys we don't have to ever add new keys, and we never modify the ending value so no need to pass it in.

### DIFF
--- a/lib/rdbms/primary_key/primary_keys.go
+++ b/lib/rdbms/primary_key/primary_keys.go
@@ -66,7 +66,7 @@ func (k *Keys) Clone() *Keys {
 func (k *Keys) UpdateStartingValue(keyName string, startingVal any) error {
 	idx := slices.IndexFunc(k.keys, func(x Key) bool { return x.Name == keyName })
 	if idx < 0 {
-		return fmt.Errorf("could not find key named %s", keyName)
+		return fmt.Errorf("no key named %s", keyName)
 	}
 
 	k.keys[idx].StartingValue = startingVal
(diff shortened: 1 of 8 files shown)


lib/rdbms/primary_key/primary_keys_test.go
```diff
@@ -12,69 +11,14 @@ func TestNewKeys(t *testing.T) {
...
-		keys.Upsert("foo", ptr.ToString("new starting value"), nil)
+		keys.UpdateStartingValue("foo", "new starting value")
 		assert.Equal(t, "foo", keys.keys[0].Name)
 		assert.Equal(t, "new starting value", keys.keys[0].StartingValue)
 		assert.Equal(t, "foo", keysArray[0].Name)
 		assert.Equal(t, 20, keysArray[0].StartingValue)
 	}
 }
 
-func TestPrimaryKeys_Length(t *testing.T) {
```
nathan-artie, line 23: `Length` wasn't being used.

lib/rdbms/primary_key/primary_keys.go
```diff
@@ -65,49 +59,18 @@ func (k *Keys) LoadValues(startingValues, endingValues []string) error {
 	return nil
 }
 
-func (k *Keys) Length() int {
```
nathan-artie, line 68: `Length` wasn't being used.

lib/rdbms/primary_key/primary_keys.go
```diff
@@ -7,19 +7,13 @@ import (
 )
 
 type Keys struct {
-	keys   []Key
-	keyMap map[string]bool
```
nathan-artie, line 11: Killed `keyMap`.

Tang8330 (approved):

PR #194, opened 2024-02-23
[mysql] Implement full MySQL driver type handling

nathan-artie:


lib/mysql/scanner/scanner.go
```diff
@@ -130,26 +132,3 @@ func (s *scanner) scan() ([]map[string]any, error) {
 
 	return rowsData, nil
 }
-
-func convertValue(colType schema.DataType, value any) (any, error) {
```
nathan-artie, line 134: Moved to https://github.com/artie-labs/reader/pull/194/files#diff-37b2e567eca56b9eda33fe49fbf65c9eff3ac2777826b3b78d22b23e0517c772R6

lib/mysql/schema/schema.go
```diff
@@ -22,7 +22,6 @@ const (
 	BigInt
 	// Fixed-Point Types (Exact Value)
 	Decimal
-	Numeric
```
nathan-artie, line 25: Columns defined as `NUMERIC` come back as `DECIMAL` in the schema so we don't need to support both. https://dev.mysql.com/doc/refman/8.0/en/fixed-point-types.html

lib/mysql/schema/schema.go
```diff
@@ -270,7 +267,8 @@ func getPrimaryKeyValues(db *sql.DB, table string, primaryKeys []Column, descend
 	if err := stmt.QueryRow().Scan(resultPtrs...); err != nil {
 		return nil, err
 	}
-	return result, nil
+
+	return ConvertValues(result, primaryKeys)
```
nathan-artie, line 271: Note added this so that we parse primary key values with the same logic we use for row values.

lib/mysql/table.go
```diff
@@ -70,13 +70,10 @@ func (t *Table) findStartAndEndPrimaryKeys(db *sql.DB) error {
 
 	t.PrimaryKeys = make([]primary_key.Key, len(primaryKeysBounds))
 	for idx, bounds := range primaryKeysBounds {
-		col := keyColumns[idx]
```
nathan-artie, line 73: Now that `primary_key.Keys` handles non-strings the min and max can stay as their native type.

lib/mysql/schema/schema.go
```diff
@@ -185,6 +180,8 @@ func parseColumnDataType(s string) (DataType, *Opts, error) {
 		return Enum, nil, nil
 	case "set":
 		return Set, nil, nil
+	case "json":
```
nathan-artie, line 183: MySQL has JSON now https://dev.mysql.com/doc/refman/8.0/en/json.html

lib/mysql/schema/convert.go
```diff
@@ -0,0 +1,103 @@
...
+			return nil, fmt.Errorf("expected []byte got %T for value: %v", value, value)
+		}
+		if len(castValue) == 0 {
+			return nil, fmt.Errorf("bit value is zero bytes: %v", value)
+		}
+		if castValue[0] == 0 {
+			return false, nil
+		} else if castValue[0] == 1 {
+			return true, nil
+		} else {
```
Tang8330, line 25: What if the length is greater than 1?

Tang8330 (approved):
🔥 🔥 🔥

PR #197, opened 2024-02-23
[postgres] Remove redundant pk `fmt.Sprint()` calls

nathan-artie:
We're already doing a `fmt.Sprint()` for primary key values here: https://github.com/artie-labs/reader/pull/197/files#diff-eb76bea08e483c31dcfdd81272705c460694b4cc6fb60a396854c96d676dd9adR155 when we build the scan query, so we don't need to do so when building and updating primary keys.

### DIFF
--- a/lib/postgres/scan.go
+++ b/lib/postgres/scan.go
@@ -73,11 +73,10 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
 		return "", err
 	}
 
-	keyNames := make([]string, len(args.PrimaryKeys))
+	quotedKeyNames := make([]string, len(args.PrimaryKeys))
 	for i, key := range args.PrimaryKeys {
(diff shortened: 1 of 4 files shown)


lib/postgres/scan.go
```diff
@@ -52,7 +52,7 @@ func (t *Table) NewScanner(db *sql.DB, primaryKeys *primary_key.Keys, batchSize
 type scanTableQueryArgs struct {
 	Schema              string
 	TableName           string
-	PrimaryKeys         *primary_key.Keys
+	PrimaryKeys         []primary_key.Key
```
nathan-artie, line 55: Also decided to pass around `[]primary_key.Key` instead of `*primary_key.Keys` since it's less mutable.

Tang8330 (approved):

PR #198, opened 2024-02-23
[postgres] Kill redundant call to `ParseValue` for pks

nathan-artie:


lib/postgres/scan.go
```diff
@@ -205,20 +205,7 @@ func _scan(s *scan.Scanner[*Table], primaryKeys *primary_key.Keys, isFirstRow bo
 	// Update the starting key so that the next scan will pick off where we last left off.
 	lastRow := rowsData[len(rowsData)-1]
 	for _, pk := range primaryKeys.Keys() {
-		col, err := s.Table.GetColumnByName(pk.Name)
-		if err != nil {
-			return nil, err
-		}
-
-		val, err := ParseValue(col.Type, ParseValueArgs{
```
nathan-artie, line 213: We already call `ParseValue` for all values here https://github.com/artie-labs/reader/pull/198/files#diff-eb76bea08e483c31dcfdd81272705c460694b4cc6fb60a396854c96d676dd9adR187
and the first thing `ParseValue` does is is check if the value has already been parsed and exit:
```
if args.Value() == nil || args.ValueWrapper.parsed {
		return args.ValueWrapper, nil
	}
```

PR #200, opened 2024-02-24
[postgres] Convert `time.Time` values to string at query time

nathan-artie:
Instead of converting primary key time.Time` values to strings when we store them in the `primary_key.Keys` we can just convert them to strings when we need to use them to build the scanning query.

### DIFF
--- a/lib/postgres/parse.go
+++ b/lib/postgres/parse.go
@@ -10,11 +10,9 @@ import (
 
 	"github.com/artie-labs/reader/lib/postgres/parse"
 	"github.com/artie-labs/reader/lib/postgres/schema"
-	"github.com/artie-labs/reader/lib/timeutil"
 )
 
 type ParseValueArgs struct {
-	ParseTime    bool
 	ValueWrapper ValueWrapper
 }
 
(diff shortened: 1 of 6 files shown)


lib/postgres/scan.go
```diff
@@ -100,6 +101,11 @@ func shouldQuoteValue(dataType schema.DataType) (bool, error) {
 }
 
 func keysToValueList(keys []primary_key.Key, columns []schema.Column) ([]string, []string, error) {
+	convertToString := func(value any) string {
+		// This is needed because we need to cast the time.Time object into a string for pagination.
+		return fmt.Sprint(timeutil.ParseValue(value))
```
Tang8330, line 106: Not related to the changes from this PR, but since you're cleaning it up - it'd be nice to rename the function name to be more descriptive

lib/postgres/parse.go
```diff
@@ -10,11 +10,9 @@ import (
 
 	"github.com/artie-labs/reader/lib/postgres/parse"
 	"github.com/artie-labs/reader/lib/postgres/schema"
-	"github.com/artie-labs/reader/lib/timeutil"
 )
 
 type ParseValueArgs struct {
```
Tang8330, line 15: Can get rid of this struct?

Tang8330 (approved):

lib/postgres/parse.go
nathan-artie: Yep, will do in a follow up PR.

PR #201, opened 2024-02-24
[mysql,postgres] Move pk updating to `Scanner`

nathan-artie:


lib/postgres/scan.go
```diff
@@ -193,44 +193,9 @@ func _scan(s *scan.Scanner[*Table], primaryKeys *primary_key.Keys, isFirstRow bo
...
-
-	// Update the starting key so that the next scan will pick off where we last left off.
-	lastRow := rowsData[len(rowsData)-1]
-	for _, pk := range primaryKeys.Keys() {
-		col, err := s.Table.GetColumnByName(pk.Name)
-		if err != nil {
-			return nil, err
-		}
-
-		val, err := ParseValue(col.Type, ParseValueArgs{
```
nathan-artie, line 213: No need to call `ParseValue` again since everything that can be parsed will have be parsed.

Tang8330 (approved):

PR #202, opened 2024-02-25
[postgres] Clean up `ParseValue`

nathan-artie:


lib/postgres/parse.go
```diff
@@ -12,109 +12,89 @@ import (
...
-		Value:  value,
-		parsed: true,
-	}
-}
-
-func ParseValue(colKind schema.DataType, args ParseValueArgs) (ValueWrapper, error) {
-	// If the value is nil, or already parsed - just return.
-	if args.Value() == nil || args.ValueWrapper.parsed {
-		return args.ValueWrapper, nil
+func ParseValue(colKind schema.DataType, value any) (any, error) {
```
Tang8330, line 15: Is it fine to get rid of NewValueWrapper? We'll lose the ability to recall this function since we no longer have `parsed` and its early return

Tang8330: And since the param for value is `any` or `interface{}`, this error will not happen during compile time, but rather run time.

nathan-artie: I think it's fine. The [two](https://github.com/artie-labs/reader/pull/202/files#diff-eb76bea08e483c31dcfdd81272705c460694b4cc6fb60a396854c96d676dd9adL187) [places](https://github.com/artie-labs/reader/pull/202/files#diff-630b53aba6f9f16efd9cd680a8fd9b7a77ff5c1a914d670a7e43893938b2af89L85) we use `ParseValue` we construct a new `NewValueWrapper` when we call it and then immediately call `.Value` on the result, so it's not currently providing any safety checks.

nathan-artie: We used to do two calls to `ParseValue` for each value, the logic was something like:
- Call `ParseValue` for each value for each row, building a mapping of key to `ValueWrapper`
- For the last row look up the value for each primary key, which will be a `ValueWrapper`, and then call `ParseValue` again on it this time also having it format times to strings.
- Loop through the original mapping of keys `ValueWrapper` for each row and now call `.Value` on each `ValueWrapper`
Now we just format time strings when constructing scan queries where we can simply check if it's a `time.Time` and then format it appropriately, so the two `ParseValue` calls end up identical and were reduced to one in https://github.com/artie-labs/reader/pull/201. Since we no longer call `ParseValue` twice for the same value the `parsed` logic is not being used.

Tang8330 (approved):

PR #209, opened 2024-02-27
Improving our MySQL source work for additional data types

Tang8330:
## Changes

1. Supporting TinyText, MediumText, LongText
2. Supporting boolean and bool
3. Consolidate Numeric decoder to use the same function as PostgreSQL


lib/debezium/converters/decimal_test.go
```diff
@@ -0,0 +1,26 @@
+package converters
+
+import (
+	"fmt"
+	"github.com/artie-labs/transfer/lib/debezium"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestDecimalConverter_Convert(t *testing.T) {
```
nathan-artie, line 10: We should add a test for `ToField` too to make sure we're adding scale and precision to the field.

lib/debezium/converters/decimal_test.go
```diff
@@ -0,0 +1,26 @@
...
+		converted, err := converter.Convert("1.23")
+		assert.NoError(t, err)
+
+		field := debezium.Field{
+			Parameters: map[string]any{
+				"scale": converter.scale,
+			},
+		}
+
+		actualValue, err := field.DecodeDecimal(fmt.Sprint(converted))
```
nathan-artie, line 22: Could do `converter.ToField("").DecodeDecimal(fmt.Sprint(converted))` here I suppose.

lib/mysql/schema/convert.go
```diff
@@ -24,6 +24,13 @@ func ConvertValue(value any, colType DataType) (any, error) {
 			return nil, fmt.Errorf("bit value is invalid: %v", value)
 		}
 		return castValue[0] == 1, nil
+	case Boolean:
```
nathan-artie, line 27: Add a test in `convert_test.go`?

PR #212, opened 2024-02-28
[postgres] Kill `TextThatRequiresEscaping`

nathan-artie:
We test all of these types in our integration test and with the exception of `inet` removing the `::text` casting produces the same output.

Removing the `::text` casting from `inet` changes the value from `192.168.1.5/32` to `192.168.1.5`.

### DIFF
--- a/lib/postgres/schema/schema.go
+++ b/lib/postgres/schema/schema.go
@@ -19,7 +19,7 @@ const (
 	Numeric
 	Bit
 	Boolean
-	TextThatRequiresCasting
+	Inet
 	Text
 	Interval
 	Array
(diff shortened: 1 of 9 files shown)


lib/postgres/schema/schema.go
```diff
@@ -147,10 +131,10 @@ func ParseColumnDataType(colKind string, precision, scale, udtName *string) (Dat
 		return Money, &Opts{
 			Scale: ptr.ToString("2"),
 		}
-	case "character varying", "text":
+	case "character varying", "text", "character", "xml", "cidr", "macaddr", "macaddr8":
 		return Text, nil
-	case "character":
-		return TextThatRequiresEscaping, nil
+	case "inet", "int4range", "int8range", "numrange", "daterange", "tsrange", "tstzrange":
```
nathan-artie, line 136: Removing the casting from`inet` changes the value from `192.168.1.5/32` to `192.168.1.5`. We don't currently test the reset of these for the integration tests so not not worrying about them right now.

Tang8330 (approved):
🔥

PR #215, opened 2024-02-28
[github] Add GitHub Action to run integration tests

nathan-artie:


.github/workflows/gha-go-test.yaml
```diff
@@ -1,5 +1,5 @@
 name: Go tests
-run-name: Running tests 🚀
+run-name: Running tests
```
nathan-artie, line 2: I feel like the 🚀 is only applicable for GitHub actions where we're deploying something.

sources/postgres/integration_test/main.go
```diff
@@ -658,7 +666,7 @@ const expectedPayloadTemplate = `{
 			"c_text": "QWERTYUIOP",
 			"c_time_with_timezone": 38057000,
 			"c_time_without_timezone": 45296000,
-			"c_timestamp_with_timezone": "2001-02-16T05:38:40-08:00",
+			"c_timestamp_with_timezone": "2001-02-16T13:38:40Z",
```
nathan-artie, line 669: The output here is actually dependent on the system timezone, so we now [set it to UTC ](https://github.com/artie-labs/reader/pull/215/files#diff-4e6bee8c22e7392833e4087c72085abeee0bd70b5f65d908c8c3bbd27edae9d7R26) so that tests work locally and with GitHub Actions.

.github/workflows/gha-integration-test.yaml
```diff
@@ -0,0 +1,25 @@
+name: Integration tests
+run-name: Running integration tests
+on: [push]
```
nathan-artie, line 3: These tests run faster than the Go tests so why not run every time?

.github/workflows/gha-go-test.yaml
Tang8330: 😂

Tang8330 (approved):

PR #219, opened 2024-02-29
[postgres] Change `castColumn` to error for unhandled types

nathan-artie:
This will ensure that we update `castColumn` when we add a new `DataType`.

I checked that `castColumn` currently works for all of the types defined here: https://github.com/artie-labs/reader/blob/60f0d745c6b8f90a49b7c374440d1be0ee5a3836/lib/postgres/schema/schema.go#L16

### DIFF
--- a/lib/postgres/cast.go
+++ b/lib/postgres/cast.go
@@ -2,20 +2,19 @@ package postgres
 
 import (
 	"fmt"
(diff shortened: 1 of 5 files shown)


Tang8330 (approved):
Yup agreed.

Explicit failure

PR #224, opened 2024-03-01
[postgres] Use value type in `convertToStringForQuery`

nathan-artie:
We should be able to depend on the Go type of `value` when we encode it for use in a query. In order to safely move towards that behavior we'll first log when we encounter a value that has a different type than the type we expect for a column and in those cases fall back to the legacy behavior. Once we've run this code for a while without error logs we'll switch to returning an error instead.

Note that `convertToStringForQuery` is only used to encode primary key values for use in the scanner query.

The longterm goal is to switch to doing a parameterized query and letting the driver take care of this string conversion.

### DIFF
--- a/lib/postgres/scanner.go
+++ b/lib/postgres/scanner.go
@@ -154,7 +154,8 @@ func convertToStringForQuery(value any, dataType schema.DataType) (string, error
 		}
 	case string:
 		switch dataType {
-		case schema.Text, schema.UserDefinedText, schema.Inet, schema.UUID:
+		case schema.Text, schema.UserDefinedText, schema.Inet, schema.UUID, schema.JSON, schema.VariableNumeric,
+			schema.Numeric, schema.Money:
 			return QuoteLiteral(fmt.Sprint(value)), nil
 		default:
 			slog.Error("string value with non-string column type",
--- a/lib/postgres/scanner.go
+++ b/lib/postgres/scanner.go
@@ -36,9 +36,22 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
 		}
 	}
 
-	startingValues, endingValues, err := keysToValueList(args.PrimaryKeys, args.Columns)
-	if err != nil {
-		return "", err
+	startingValues := make([]string, len(args.PrimaryKeys))
+	endingValues := make([]string, len(args.PrimaryKeys))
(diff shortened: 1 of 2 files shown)


lib/postgres/scanner.go
```diff
@@ -36,9 +36,22 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
 		}
 	}
 
-	startingValues, endingValues, err := keysToValueList(args.PrimaryKeys, args.Columns)
-	if err != nil {
-		return "", err
+	startingValues := make([]string, len(args.PrimaryKeys))
```
nathan-artie, line 39: Inlined `keysToValueList`.

lib/postgres/scanner.go
```diff
@@ -97,46 +111,72 @@ func shouldQuoteValue(dataType schema.DataType) (bool, error) {
...
 	}
 }
 
-func keysToValueList(keys []primary_key.Key, columns []schema.Column) ([]string, []string, error) {
-	convertToString := func(value any) string {
-		switch castValue := value.(type) {
-		case time.Time:
-			// This is needed because we need to cast the time.Time object into a string for pagination.
-			return castValue.Format(time.RFC3339)
+func convertToStringForQuery(value any, dataType schema.DataType) (string, error) {
```
nathan-artie, line 118: We should be able to depend on the Go type of `value` when we stringify a value for use in a query. In order to safely move towards that behavior we'll first log when we encounter a value that has a different type than the type we expect for a column and in those cases fall back to the legacy behavior.

lib/postgres/scanner.go
```diff
@@ -118,19 +118,67 @@ func shouldQuoteValue(dataType schema.DataType) (bool, error) {
...
 			return fmt.Sprint(value), nil
+		default:
+			slog.Error("bool value with non-bool column type",
+				slog.Any("value", value),
+				slog.Any("dataType", dataType),
+			)
 		}
+	case string:
+		switch dataType {
+		case schema.Text, schema.UserDefinedText, schema.Inet, schema.UUID, schema.JSON, schema.VariableNumeric,
```
nathan-artie, line 157: Note that `shouldQuoteValue` returns true for all of these types.

Tang8330 (approved):

PR #232, opened 2024-03-06
[mysql,postgres] Only store primary key names on `Table`s

nathan-artie:


lib/rdbms/scan/scan.go
```diff
@@ -14,11 +14,6 @@ const (
 	jitterMaxMs  = 5000
 )
 
-type Table interface {
```
nathan-artie, line 17: Unused.

Tang8330 (approved):

PR #236, opened 2024-03-06
[debezium] Add `NewIterator()` to `Adapter` interface

nathan-artie:


lib/debezium/transformer.go
```diff
@@ -11,21 +11,29 @@ import (
 	"github.com/artie-labs/reader/lib"
 )
 
+type Row = map[string]any
```
Tang8330, line 14:
```go
type Row map[string]any
``` 

instead?

nathan-artie: Are they equivalent or does that doe something different?

nathan-artie: Hmm that doesn't seem to work.

Tang8330: oh you are trying to do an alias, I see.

Tang8330 (approved):

PR #239, opened 2024-03-07
[postgres] Stricter column type handling

nathan-artie:
Kill `InvalidDataType` and instead throw an error in `DescribeTable` if we encounter a column type we don't support.

### DIFF
--- a/sources/postgres/adapter/adapter.go
+++ b/sources/postgres/adapter/adapter.go
@@ -34,7 +34,7 @@ func NewPostgresAdapter(db *sql.DB, tableCfg config.PostgreSQLTable) (postgresAd
 	for i, col := range table.Columns {
 		fields[i], err = ColumnToField(col)
 		if err != nil {
-			return postgresAdapter{}, err
+			return postgresAdapter{}, fmt.Errorf("failed to build field for column %s: %w", col.Name, err)
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


sources/postgres/adapter/adapter_test.go
```diff
@@ -47,26 +45,6 @@ func TestPostgresAdapter_TopicSuffix(t *testing.T) {
 	}
 }
 
-func TestPostgresAdapter_Fields(t *testing.T) {
```
nathan-artie, line 50: Killed this test since it's not easy to test anymore.

sources/postgres/adapter/adapter.go
```diff
@@ -29,9 +30,18 @@ func NewPostgresAdapter(db *sql.DB, tableCfg config.PostgreSQLTable) (postgresAd
 		return postgresAdapter{}, fmt.Errorf("failed to load metadata for table %s.%s: %w", tableCfg.Schema, tableCfg.Name, err)
 	}
 
+	fields := make([]transferDbz.Field, len(table.Columns))
```
nathan-artie, line 33: Moved this here so we can handle errors when we're creating an adapter and not when we're trying to build Debezium messages.

sources/postgres/adapter/fields.go
```diff
@@ -11,103 +13,106 @@ type Result struct {
...
 		return Result{
 			DebeziumType: string(debezium.Timestamp),
 			// NOTE: We are returning string here because we want the right layout to be used by our Typing library
 			Type: "string",
-		}
+		}, nil
 	}
 
-	return Result{}
+	return Result{}, fmt.Errorf("unsupported data type DataType(%d)", d)
```
nathan-artie, line 108: Checked that all of our `DataType`s are handled in this function.

Tang8330 (approved):

PR #246, opened 2024-03-08
[postgres] Add `VariableNumericConverter`

nathan-artie:
Will merge after https://github.com/artie-labs/reader/pull/247 since that PR adds a test that will ensure there aren't any data conversion changes in this PR.

### DIFF
--- a/lib/debezium/converters/decimal.go
+++ b/lib/debezium/converters/decimal.go
@@ -53,8 +53,8 @@ func (VariableNumericConverter) ToField(name string) transferDBZ.Field {
 }
 
 func (VariableNumericConverter) Convert(value any) (any, error) {
-	stringValue, isOk := value.(string)
-	if !isOk {
+	stringValue, ok := value.(string)
+	if !ok {
 		return nil, fmt.Errorf("expected string got %T with value: %v", value, value)
 	}
 
--- a/lib/debezium/converters/decimal.go
+++ b/lib/debezium/converters/decimal.go
@@ -2,6 +2,7 @@ package converters
 
 import (
 	"fmt"
+
 	transferDBZ "github.com/artie-labs/transfer/lib/debezium"
 
 	"github.com/artie-labs/reader/lib/debezium"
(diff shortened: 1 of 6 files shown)


lib/debezium/converters/decimal.go
```diff
@@ -40,3 +41,32 @@ func (d decimalConverter) Convert(value any) (any, error) {
...
+func (VariableNumericConverter) ToField(name string) transferDBZ.Field {
+	return transferDBZ.Field{
+		FieldName:    name,
+		Type:         "struct",
+		DebeziumType: string(transferDBZ.KafkaVariableNumericType),
+	}
+}
+
+func (VariableNumericConverter) Convert(value any) (any, error) {
+	stringValue, isOk := value.(string)
```
nathan-artie, line 56: This should be safe for Postgres because when we read `VariableNumeric` values we' throw an error if they aren't strings: https://github.com/artie-labs/reader/blob/4ae589472e29e65219113b28fa0655e8ea68f817/lib/postgres/parse/parse.go#L61

lib/debezium/converters/decimal.go
```diff
@@ -40,3 +41,32 @@ func (d decimalConverter) Convert(value any) (any, error) {
...
+func (VariableNumericConverter) ToField(name string) transferDBZ.Field {
+	return transferDBZ.Field{
+		FieldName:    name,
+		Type:         "struct",
+		DebeziumType: string(transferDBZ.KafkaVariableNumericType),
+	}
+}
+
+func (VariableNumericConverter) Convert(value any) (any, error) {
+	stringValue, ok := value.(string)
```
nathan-artie, line 56: This should be safe for Postgres because when we read `VariableNumeric` values we return an error if they aren't strings: https://github.com/artie-labs/reader/blob/4ae589472e29e65219113b28fa0655e8ea68f817/lib/postgres/parse/parse.go#L61

Tang8330 (approved):

PR #248, opened 2024-03-08
[postgres] Add `MoneyConverter`

nathan-artie:


lib/postgres/schema/schema.go
```diff
@@ -132,9 +131,7 @@ func ParseColumnDataType(colKind string, precision, scale, udtName *string) (Dat
 	case "time with time zone", "time without time zone":
 		return Time, nil, nil
 	case "money":
-		return Money, &Opts{
-			Scale: ptr.ToString("2"),
```
nathan-artie, line 136: Since the scale is always `2` I just moved it to the converter.

Tang8330 (approved):

PR #250, opened 2024-03-08
[postgres] Use `DecimalConverter` for numeric columns

nathan-artie:


lib/postgres/schema/schema.go
```diff
@@ -69,8 +69,8 @@ func DescribeTable(db *sql.DB, _schema, table string) ([]Column, error) {
 	for rows.Next() {
 		var colName string
 		var colType string
-		var numericPrecision *string
-		var numericScale *string
+		var numericPrecision *int
```
nathan-artie, line 72: Checked the schema and these two are defined as `cardinal_number` https://www.postgresql.org/docs/current/infoschema-datatypes.html#INFOSCHEMA-DATATYPES

lib/postgres/schema/schema.go
```diff
@@ -146,8 +146,12 @@ func ParseColumnDataType(colKind string, precision, scale, udtName *string) (Dat
 			if precision == nil && scale == nil {
 				return VariableNumeric, nil, nil
 			} else {
+				if scale == nil {
```
nathan-artie, line 149: @Tang8330 Do you know if it's possible for only one of these to be nil, or should they both either be nil or non-nil?

sources/postgres/adapter/values.go
```diff
@@ -30,16 +27,6 @@ func ConvertValueToDebezium(col schema.Column, value any) (any, error) {
 				return nil, nil
 			}
 		}
-	case schema.Numeric:
```
nathan-artie, line 33: Compare to https://github.com/artie-labs/reader/blob/07d0ff46290bfbed0ada3e67f31c7cd9b48bed67/lib/debezium/converters/decimal.go#L36

lib/postgres/schema/schema.go
Tang8330: I don't think they would ever be both nil

https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL

sources/postgres/adapter/values.go
nathan-artie: Note that we ensure that `numeric` values are `string`s when we parse them: https://github.com/artie-labs/reader/blob/07d0ff46290bfbed0ada3e67f31c7cd9b48bed67/lib/postgres/parse/parse.go#L61

lib/postgres/schema/schema.go
Tang8330: * Numeric(p,s) - both are set
* Numeric(p) - scale is set on an ad-hoc basis (Variable Numeric)
* Numeric - scale is implied 0

sources/postgres/adapter/fields.go
```diff
@@ -114,15 +114,7 @@ func ColumnToField(col schema.Column) (debezium.Field, error) {
 	}
 
 	if col.Opts != nil {
```
Tang8330, line 116: If this the case, can we just kill opts on col?

Tang8330 (approved):

lib/postgres/schema/schema.go
```diff
@@ -145,13 +145,13 @@ func ParseColumnDataType(colKind string, precision, scale *int, udtName *string)
...
-			} else if scale != nil {
+			} else if precision != nil && scale != nil {
 				return Numeric, &Opts{
 					Scale:     *scale,
-					Precision: precision,
+					Precision: *precision,
 				}, nil
 			} else {
-				return -1, nil, fmt.Errorf("scale is nil but precision is not")
+				return -1, nil, fmt.Errorf("expected precision (%v) and scale (%v) to both be nil or not-nil", precision, scale)
```
Tang8330, line 154: This is not going to print well if the value is actually set since it's a pointer, it'll just show the memory address

sources/postgres/adapter/fields.go
nathan-artie: This whole function will be going away when everything has a value converter, I just left this statement as a a sanity check for now.

Tang8330: cool

lib/postgres/schema/schema.go
nathan-artie: Ooh good catch!

PR #251, opened 2024-03-08
[postgres] Use `JSONConverter` for json types

nathan-artie:


sources/postgres/adapter/fields.go
```diff
@@ -72,11 +72,6 @@ func toDebeziumType(d schema.DataType) (Result, error) {
 			DebeziumType: "io.debezium.data.Uuid",
 			Type:         "string",
 		}, nil
-	case schema.JSON:
```
nathan-artie, line 75: Compare to https://github.com/artie-labs/reader/blob/9a185de9c4bcc412a4d9022bdfd02db6bae30a0e/lib/debezium/converters/json.go#L12

nathan-artie: `debezium.JSON` = `io.debezium.data.Json`

sources/postgres/adapter/adapter.go
```diff
@@ -98,12 +98,14 @@ func valueConverterForType(dataType schema.DataType, opts *schema.Opts) converte
...
+	case schema.Money:
+		return MoneyConverter{}
 	case schema.Bytea:
 		return converters.BytesPassthrough{}
 	case schema.Date:
 		return converters.DateConverter{}
-	case schema.Money:
-		return MoneyConverter{}
+	case schema.JSON:
+		return converters.JSONConverter{}
```
nathan-artie, line 108: Note that we convert the `[]byte` value we receive from the driver to `string` when we parse it https://github.com/artie-labs/reader/blob/9a185de9c4bcc412a4d9022bdfd02db6bae30a0e/lib/postgres/parse/parse.go#L60 so the check for a string type `JSONConverter` is fine.

Tang8330 (approved):

PR #254, opened 2024-03-08
[postgers] Add `MapConverter` for `hstore`

nathan-artie:


lib/debezium/converters/map.go
```diff
@@ -0,0 +1,25 @@
...
+
+func (MapConverter) ToField(name string) debezium.Field {
+	return debezium.Field{
+		FieldName: name,
+		Type:      "map",
+	}
+}
+
+func (MapConverter) Convert(value any) (any, error) {
+	mapValue, ok := value.(map[string]any)
```
nathan-artie, line 19: For `hstore` we always emit a `map[string]any`: https://github.com/artie-labs/reader/blob/42a3420ccde9d1d0310005ee8bfa3622ec0b0f3d/lib/postgres/parse/parse.go#L106

Tang8330 (approved):

PR #255, opened 2024-03-08
[debezium,postgres] Add geometry value converters

nathan-artie:


lib/debezium/converters/geometry.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+}
+
+func (_geomConverter) Convert(value any) (any, error) {
+	mapValue, ok := value.(map[string]any)
+	if !ok {
+		return nil, fmt.Errorf("expected map[string]any got %T with value: %v", value, value)
+	}
+
+	return mapValue, nil
```
nathan-artie, line 27: We're emit `map[string]any` for all three types:
- https://github.com/artie-labs/reader/blob/42a3420ccde9d1d0310005ee8bfa3622ec0b0f3d/lib/postgres/parse/parse.go#L32
- https://github.com/artie-labs/reader/blob/42a3420ccde9d1d0310005ee8bfa3622ec0b0f3d/lib/postgres/parse/parse.go#L44

PR #257, opened 2024-03-08
[postgres] Add passthrough converter for rest of types

nathan-artie:
Adding a temporary converter for postgres that allows values to go through without checking the type. This way we can switch to using the `RowConverter` for the Postgres adapter which is what we're using for the MySQL adapter.


sources/postgres/adapter/adapter.go
```diff
@@ -74,63 +79,55 @@ func (p postgresAdapter) PartitionKey(row map[string]any) map[string]any {
...
 	case schema.Point:
-		return converters.NewPointConverter()
+		return converters.NewPointConverter(), nil
 	case schema.Geometry:
-		return converters.NewGeometryConverter()
+		return converters.NewGeometryConverter(), nil
 	case schema.Geography:
-		return converters.NewGeographyConverter()
+		return converters.NewGeographyConverter(), nil
+	case schema.Boolean, schema.Bit:
```
nathan-artie, line 112: Compare the following with https://github.com/artie-labs/reader/pull/257/files#diff-36e69de1416f8376e1d364aba58764c18f0f7662b95627bb0fb2301e63162aadL18

sources/postgres/adapter/fields_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/artie-labs/reader/lib/postgres/schema"
 )
 
-func TestColumnToField(t *testing.T) {
+func TestValueConverterForType_Fields(t *testing.T) {
```
nathan-artie, line 12: Leaving this here for now so that the diff is smaller, but will move this test to `adapter_test.go` since `fields.go` is no more.

Tang8330 (approved):

PR #261, opened 2024-03-09
[MySQL] Adding more logging so we can see the row that failed 

Tang8330:
Logging out the entire row so we can debug what is wrong


Tang8330:
Added a card for this instead so we can see come up with a better approach

PR #262, opened 2024-03-09
Skip MySQL Zero Dates

Tang8330:
We were previously failing on this, however this is valid in MySQL.

> The "0000-00-00" date is often used in MySQL databases as a placeholder for an unknown or default date. However, this value is not considered a valid date in many contexts, including some programming languages or database drivers that might be used to interact with MySQL.

### DIFF
--- a/lib/mysql/schema/convert.go
+++ b/lib/mysql/schema/convert.go
@@ -71,6 +71,13 @@ func ConvertValue(value any, colType DataType) (any, error) {
 		if !ok {
 			return nil, fmt.Errorf("expected []byte got %T for value: %v", value, value)
 		}
+
+		// TODO: We should check for zero datetime, timestamp, etc.
+		// MySQL returns 0000-00-00 for zero dates
+		if string(bytesValue) == "0000-00-00" {
+			return nil, nil
+		}
+
 		timeValue, err := time.Parse(time.DateOnly, string(bytesValue))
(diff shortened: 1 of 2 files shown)


lib/mysql/schema/convert.go
```diff
@@ -123,7 +130,7 @@ func ConvertValues(values []any, cols []Column) ([]any, error) {
 		col := cols[idx]
 		convertedVal, err := ConvertValue(value, col.Type)
 		if err != nil {
-			return nil, fmt.Errorf("faild to convert value for column %s: %w", col.Name, err)
+			return nil, fmt.Errorf("failed to convert value for column %s: %w", col.Name, err)
```
nathan-artie, line 133: Lol

lib/mysql/schema/convert.go
```diff
@@ -71,6 +71,13 @@ func ConvertValue(value any, colType DataType) (any, error) {
 		if !ok {
 			return nil, fmt.Errorf("expected []byte got %T for value: %v", value, value)
 		}
+
+		// TODO: We should check for zero datetime, timestamp, etc.
+		// MySQL returns 0000-00-00 for zero dates
+		if string(bytesValue) == "0000-00-00" {
```
nathan-artie, line 77: Add a test for this case?

PR #263, opened 2024-03-10
[mysql] Move handling of `0000-00-00` dates to `DateConverter`

nathan-artie:


lib/debezium/converters/time.go
```diff
@@ -46,13 +46,27 @@ func (DateConverter) ToField(name string) debezium.Field {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse string '%s' to date: %w", castValue, err)
+		}
+	default:
+		return nil, fmt.Errorf("expected string/time.Time got %T with value: %v", value, value)
 	}
 
 	unix := time.UnixMilli(0).In(time.UTC) // 1970-01-01
-	return int(ts.Sub(unix).Hours() / 24), nil
+	return int32(timeValue.Sub(unix).Hours() / 24), nil
```
nathan-artie, line 69: Might as well use `int32` since that's what we have here https://github.com/artie-labs/reader/pull/263/files#diff-5de60c74c1cf57a3f6e6d17dc239b2e77d2e9b9ac954257b55972fdc22926f0cL43

lib/debezium/converters/time.go
```diff
@@ -46,13 +46,27 @@ func (DateConverter) ToField(name string) debezium.Field {
...
-	if !isOk {
-		return nil, fmt.Errorf("object is not a time.Time object")
+	var timeValue time.Time
+	switch castValue := value.(type) {
+	case time.Time:
+		timeValue = castValue
+	case string:
+		if castValue == "0000-00-00" {
+			// MySQL supports '0000-00-00' for date columns
+			// TODO: We should check for zero datetime, timestamp, etc.
```
Tang8330, line 56: Remove this TODO since it's no longer relevant to DateConverter?

lib/debezium/converters/time.go
```diff
@@ -46,13 +46,27 @@ func (DateConverter) ToField(name string) debezium.Field {
...
+			return nil, nil
+		}
+		var err error
+		timeValue, err = time.Parse(time.DateOnly, castValue)
```
Tang8330, line 60: Can we also check that year doesn't exceed 9999 or below 0 as well?

Tang8330 (approved):
lgtm % c

lib/debezium/converters/time.go
nathan-artie: We can, but unlike `TimestampConverter` where we output a `time.Time` here we're outputting the number of hours since 1970 so if the date is above 9999 or below 0 it won't cause a JSON Marshal error.

nathan-artie: Hah `time.Duration` can't represent durations > 290 years, so this doesn't even work for years > 2260. I'll address in a follow up PR.

PR #265, opened 2024-03-10
[debezium] Fix `DateConverter` for dates > 290 years from 1970

nathan-artie:
`time.Duration` [doesn't work for durations > 290 years](https://pkg.go.dev/time#pkg-types).
> A Duration represents the elapsed time between two instants as an int64 nanosecond count. The representation limits the largest representable duration to approximately 290 years.

### DIFF
--- a/lib/debezium/converters/time_test.go
+++ b/lib/debezium/converters/time_test.go
@@ -66,16 +66,58 @@ func TestDateConverter_Convert(t *testing.T) {
 		assert.Equal(t, int32(19480), value)
 	}
 	{
-		// time.Time
+		// time.Time - Unix epoch
+		days, err := converter.Convert(time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC))
+		assert.NoError(t, err)
+		assert.Equal(t, int32(0), days)
+	}
+	{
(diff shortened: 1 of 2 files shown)


lib/debezium/converters/time_test.go
```diff
@@ -66,17 +66,59 @@ func TestDateConverter_Convert(t *testing.T) {
...
+		assert.Equal(t, int32(-365), days)
+	}
+	{
+		// time.Time - Year 9999
+		days, err := converter.Convert(time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC))
+		assert.NoError(t, err)
+		assert.Equal(t, int32(2_932_532), days)
+	}
+	{
+		// time.Time - Year 10_000
```
nathan-artie, line 105: @Tang8330 As we're outputting an `int32` we won't have a JSON Marshal error like we did for `TimestampConverter`, will it still cause an issue with Debezium if the year is > 9999?

Tang8330: If it's year specifically, it's fine since it doesn't require usage of `time.Time`

Tang8330 (approved):

PR #269, opened 2024-03-10
[postgres] Add `ArrayConverter`

nathan-artie:


lib/postgres/parse/parse.go
```diff
@@ -66,13 +65,20 @@ func ParseValue(colKind schema.DataType, value any) (any, error) {
 
 		return nil, fmt.Errorf("value: %v not of string type for Numeric or VariableNumeric", value)
 	case schema.Array:
-		var arr []any
-		if reflect.TypeOf(value).Kind() == reflect.Slice {
```
nathan-artie, line 70: `reflect.TypeOf(value).Kind() == reflect.Slice` will match `[]byte` which is undesirable.

lib/postgres/parse/parse.go
```diff
@@ -66,13 +65,20 @@ func ParseValue(colKind schema.DataType, value any) (any, error) {
...
-			// If it's already a slice, don't modify it further.
-			return value, nil
+		var bytesValue []byte
+		switch castValue := value.(type) {
+		case []byte:
+			bytesValue = castValue
+		case string:
```
nathan-artie, line 72: In practice the value will always be a `string` since we're casting it here https://github.com/artie-labs/reader/blob/608ff8404cf6c8bb2dc63acbe3cd0eeedba6306f/lib/postgres/cast.go#L29

Tang8330: So is there a point in adding `case []byte`?

Tang8330 (approved):

lib/postgres/parse/parse.go
nathan-artie: Eh not really, we can probably kill it.

PR #273, opened 2024-03-11
[debezium] Check for negative overflow when coercing ints

nathan-artie:


Tang8330 (approved):

lib/debezium/converters/int.go
```diff
@@ -10,17 +10,17 @@ func asInt16(value any) (int16, error) {
 	case int16:
 		return castValue, nil
 	case int32:
-		if castValue > math.MaxInt16 {
+		if castValue > math.MaxInt16 || castValue < math.MinInt16 {
 			return 0, fmt.Errorf("value is too large for int16")
```
Tang8330, line 14: too small?

Tang8330: sgtm

PR #274, opened 2024-03-11
[postgres] Add `PgTimeConverter`

nathan-artie:
- No longer need to cast `time without time zone` columns
- Still have to cast `time with time zone` but now it's just converting to `UTC` rather than Unix epoch
- We now support using `time without time zone` as a primary key 
  - `time without time zone` is harder to support as `pgtype.Time` doesn't parse them properly: https://github.com/jackc/pgx/issues/1940

### DIFF
--- a/lib/postgres/cast.go
+++ b/lib/postgres/cast.go
@@ -23,7 +23,7 @@ func castColumn(col schema.Column) (string, error) {
 		schema.UserDefinedText, schema.Text,
 		schema.Money, schema.VariableNumeric, schema.Numeric,
 		schema.Boolean, schema.Bit, schema.Bytea,
-		schema.TimeWithoutTimeZone, schema.Date, schema.Timestamp, schema.Interval, schema.HStore, schema.JSON,
+		schema.Time, schema.Date, schema.Timestamp, schema.Interval, schema.HStore, schema.JSON,
 		schema.Point, schema.Geography, schema.Geometry:
 		// These are all the columns that do not need to be escaped.
 		return colName, nil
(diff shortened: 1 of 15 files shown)


lib/postgres/cast.go
```diff
@@ -13,19 +13,18 @@ func castColumn(col schema.Column) (string, error) {
...
-		// We want to extract(epoch from interval) will emit this in ms
-		// However, Debezium wants this in microseconds, so we are multiplying this by 1000.
-		// We need to use CAST, because regular ::int makes this into a bytes array.
-		// extract from epoch outputs in seconds, default multiplier to ms.
-		return fmt.Sprintf(`cast(extract(epoch from %s)*%d as bigint) as "%s"`, colName, 1000, col.Name), nil
+	case schema.TimeWithTimeZone:
+		// If we don't convert `time with time zone` to UTC we end up with strings like `10:23:54-02`
+		// And pgtype.Time doesn't parse the offset propertly.
+		// See https://github.com/jackc/pgx/issues/1940
+		return fmt.Sprintf(`%s AT TIME ZONE 'UTC' AS "%s"`, colName, col.Name), nil
```
Tang8330, line 20: I wonder if we can set this within our SQL session

Tang8330 (approved):

lib/postgres/cast.go
nathan-artie: That doesn't seem to change the output:
```
postgres=# set time zone 'utc';
SET
postgres=# select time_col from foo;
  time_col   
-------------
 05:34:17-05
 01:34:17-04
(2 rows)
```

Tang8330: Cool, good to know

PR #275, opened 2024-03-11
[postgres] Check primary key data types before scanning

nathan-artie:


lib/postgres/schema/schema.go
```diff
@@ -14,28 +14,28 @@ import (
 type DataType int
 
 const (
-	VariableNumeric DataType = iota
-	Money
+	Int16 DataType = iota
```
nathan-artie, line 17: Just reordered these.

Tang8330 (approved):

PR #279, opened 2024-03-12
[debezium] Kill `Keys.Clone()`

nathan-artie:


lib/rdbms/scan/scan.go
```diff
@@ -56,7 +56,7 @@ func NewScanner(db *sql.DB, _primaryKeys []primary_key.Key, cfg ScannerConfig, a
 		batchSize:    cfg.BatchSize,
 		retryCfg:     retryCfg,
 		adapter:      adapter,
-		primaryKeys:  primaryKeys.Clone(),
+		primaryKeys:  primaryKeys,
```
nathan-artie, line 59: Since `primaryKeys` are now initialized within this function we don't have to worry about cloning them anymore.

Tang8330 (approved):

PR #282, opened 2024-03-12
[mysql,postgres] Offset `DataType`s by one

nathan-artie:
This way if we forget to pass a `DataType` in to a struct it'll default to the `0` value which is invalid and we'll get an error in our switch statements rather than `Bit` or `TinyInt`.

### DIFF
--- a/lib/postgres/schema/schema.go
+++ b/lib/postgres/schema/schema.go
@@ -14,7 +14,7 @@ import (
 type DataType int
 
 const (
-	Bit DataType = iota
+	Bit DataType = iota + 1
 	Boolean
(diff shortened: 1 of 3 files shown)


lib/postgres/schema/schema.go
```diff
@@ -14,7 +14,7 @@ import (
 type DataType int
 
 const (
-	Bit DataType = iota
+	Bit DataType = iota + 1
```
Tang8330, line 17: Yeah, that's why whenever I use iota, I make 0 invalid.

Tang8330 (approved):

PR #284, opened 2024-03-13
[postgres] Kill `shouldQuoteValue`

nathan-artie:
Instead of quoting a column based on its `DataType` we'll just quote it if it's a string and otherwise not quote it.

### DIFF
--- a/lib/postgres/scanner.go
+++ b/lib/postgres/scanner.go
@@ -3,7 +3,6 @@ package postgres
 import (
 	"database/sql"
 	"fmt"
-	"log/slog"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


lib/postgres/scanner.go
```diff
@@ -176,30 +144,6 @@ func convertToStringForQuery(value any, dataType schema.DataType) (string, error
...
-		case schema.Text, schema.UserDefinedText, schema.Inet, schema.UUID, schema.JSON, schema.VariableNumeric,
-			schema.Numeric, schema.Money:
-			return QuoteLiteral(castValue), nil
-		default:
-			slog.Error("string value with non-string column type",
-				slog.String("value", castValue),
-				slog.Any("dataType", dataType),
-			)
-			// legacy behavior - used when optionalPrimaryKeyValStart/End is configured
-			// TODO: parse optionalPrimaryKeyValStart/End based on DataType to Go type
```
nathan-artie, line 192: Added this in https://github.com/artie-labs/reader/pull/281

Tang8330 (approved):

PR #285, opened 2024-03-13
[rdbms] Add generic `Column` type

nathan-artie:


lib/postgres/parse/parse.go
```diff
@@ -89,7 +89,6 @@ func ParseValue(colKind schema.DataType, value any) (any, error) {
 
 		return _uuid.String(), nil
 	case schema.JSON:
-		// Debezium sends JSON as a JSON string
```
nathan-artie, line 92: Isn't relevent here.

Tang8330 (approved):

PR #290, opened 2024-03-14
[postgres] Use parameterized scanning query

nathan-artie:


lib/postgres/scanner.go
```diff
@@ -56,30 +56,25 @@ type scanTableQueryArgs struct {
...
+			return "", nil, err
 		}
 	}
 
-	startingValues := make([]string, len(args.PrimaryKeys))
-	endingValues := make([]string, len(args.PrimaryKeys))
+	startingValues := make([]any, len(args.PrimaryKeys))
+	endingValues := make([]any, len(args.PrimaryKeys))
 	for i, pk := range args.PrimaryKeys {
-		colIndex := slices.IndexFunc(args.Columns, func(col schema.Column) bool { return col.Name == pk.Name })
```
nathan-artie, line 72: Code isn't necessary anymore, `colIndex` isn't actually used.

lib/postgres/scanner.go
```diff
@@ -99,52 +99,24 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
...
 	switch castValue := value.(type) {
-	case bool, int, int8, int16, int32, int64, float32, float64:
-		return fmt.Sprint(value), nil
-	case string:
-		return QuoteLiteral(castValue), nil
+	case nil, bool, int, int8, int16, int32, int64, float32, float64, string, pgtype.Time, pgtype.Interval:
+		return value, nil
 	case time.Time:
-		return QuoteLiteral(castValue.Format(time.RFC3339)), nil
-	case pgtype.Time:
```
nathan-artie, line 122: `pgtype.Time` and `pgtype.Interval` both implement the `database/sql/driver` `Valuer` interface which means that when they are used as a parameter in a query the `.Value()` method will be called and then the result encoded (in this case they will both be strings so they will be quoted) and passed along to the driver.

lib/postgres/scanner.go
```diff
@@ -99,52 +99,24 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
...
-		if err != nil {
-			return "", err
-		}
-		stringValue, ok := value.(string)
-		if !ok {
-			return "", fmt.Errorf("expected string got %T with value %v", value, value)
-		}
-		return QuoteLiteral(stringValue), nil
+		// TODO: See if we can directly use `time.Time` in query
+		return castValue.Format(time.RFC3339), nil
```
nathan-artie, line 119: We should be able to kill this method entirely but I need to do some research into how `time.Time` is encoded when used as a query parameter.

nathan-artie: After more research I think it should be fine, the existing code actually has a bug in that we're throwing away nanoseconds when we format times (using `time.RFC3339` instead of `time.RFC3339Nano`) which means it wouldn't work for times that have fractional seconds in them.

integration_tests/postgres/main.go
```diff
@@ -937,9 +937,9 @@ INSERT INTO %s VALUES (
 	-- c_time_without_timezone
 		'20:38:21',
 	-- c_timestamp_without_timezone
-		'2001-02-16 20:38:40',
+		'2001-02-16 20:38:40.123123',
```
nathan-artie, line 940: This would have failed before as it has fractional seconds.

lib/postgres/scanner.go
```diff
@@ -99,55 +92,14 @@ func scanTableQuery(args scanTableQueryArgs) (string, error) {
...
-
-// convertToStringForQuery returns a string value suitable for use directly in a query.
-func convertToStringForQuery(value any) (string, error) {
-	// TODO: Switch to using a parameterized query
-	switch castValue := value.(type) {
-	case bool, int, int8, int16, int32, int64, float32, float64:
-		return fmt.Sprint(value), nil
-	case string:
-		return QuoteLiteral(castValue), nil
-	case time.Time:
```
nathan-artie, line 120: There's actually a small bug here in that we're throwing away nanoseconds when we format times (using time.RFC3339 instead of time.RFC3339Nano) which means could end up missing rows with the scanner if the times have fractional seconds in them.

Tang8330 (approved):
Nice!

PR #292, opened 2024-03-14
[postgres] Support `bytea` primary keys

nathan-artie:


lib/rdbms/primary_key/primary_keys.go
```diff
@@ -84,9 +85,23 @@ func (k *Keys) Keys() []Key {
...
 		}
 	}
 	return true
 }
+
+func equal(a, b any) bool {
+	if aBytes, ok := a.([]byte); ok {
+		bBytes, ok := b.([]byte)
+		if !ok {
+			return false
```
Tang8330, line 99: Is there something special about byte array when checking for equality?

If so, let's leave a comment?

Tang8330 (approved):
Lgtm

lib/rdbms/primary_key/primary_keys.go
nathan-artie: Yeah you can't compare two of them: `comparing uncomparable type []uint8`

PR #293, opened 2024-03-14
[postgres] Support `bit` primary keys

nathan-artie:


lib/debezium/converters/bit.go
```diff
@@ -0,0 +1,30 @@
...
+		FieldName: name,
+		Type:      "boolean",
+	}
+}
+
+func (BitConverter) Convert(value any) (any, error) {
+	// This will be 0 (false) or 1 (true)
+	stringValue, ok := value.(string)
+	if ok {
+		if stringValue == "0" {
```
Tang8330, line 22: nit: can do

```
return stringValue == "1"
```

Tang8330 (approved):

PR #299, opened 2024-03-17
[postgres] Default to not casting columns

nathan-artie:
Since casting is the exception we don't need to be explicit about which columns are not cast.


Tang8330 (approved):
sgtm

PR #302, opened 2024-03-17
[debezium] Return `[]byte` from decimal converters

nathan-artie:
When a `[]byte` is marshalled to JSON [it'll be base64 encoded](https://github.com/golang/go/blob/0a6f05e30f58023bf45f747a79c20751db2bcfe7/src/encoding/json/encode.go#L781), so our value converters can just return byte arrays.

### DIFF
--- a/integration_tests/postgres/main.go
+++ b/integration_tests/postgres/main.go
@@ -679,7 +679,7 @@ const expectedPayloadTemplate = `{
 			"c_money": "T30t",
 			"c_numeric": "AYHN",
 			"c_numeric_variable": {
-				"scale": "5",
+				"scale": 5,
 				"value": "QX3UWQ=="
 			},
 			"c_numrange": "[11.1,22.2)",
--- a/lib/debezium/converters/decimal.go
+++ b/lib/debezium/converters/decimal.go
@@ -64,7 +64,6 @@ func (VariableNumericConverter) Convert(value any) (any, error) {
 	}
 
 	scale := debezium.GetScale(stringValue)
-
 	return VariableScaleDecimal{
 		Scale: int32(scale),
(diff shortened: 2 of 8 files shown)


integration_tests/postgres/main.go
```diff
@@ -679,7 +679,7 @@ const expectedPayloadTemplate = `{
 			"c_money": "T30t",
 			"c_numeric": "AYHN",
 			"c_numeric_variable": {
-				"scale": "5",
+				"scale": 5,
```
nathan-artie, line 682: Checked this against Debezium output and `scale` is actually an `int32`: https://debezium.io/documentation/reference/stable/connectors/postgresql.html#:~:text=debezium.data.VariableScaleDecimal-,Contains%20a%20structure%20with%20two%20fields%3A%20scale%20of%20type%20INT32%20that%20contains%20the%20scale%20of%20the%20transferred%20value%20and%20value%20of%20type%20BYTES%20containing%20the%20original%20value%20in%20an%20unscaled%20form.,-DECIMAL

lib/debezium/converters/decimal.go
```diff
@@ -39,8 +39,7 @@ func (d decimalConverter) Convert(value any) (any, error) {
 	if !isOk {
 		return nil, fmt.Errorf("expected string got %T with value: %v", value, value)
 	}
-
-	return debezium.EncodeDecimalToBase64(castValue, d.scale)
+	return debezium.EncodeDecimalToBytes(castValue, d.scale), nil
```
nathan-artie, line 42: Since `Type:         "bytes",` for the field I think it makes sense for this to be a byte array.

Tang8330 (approved):

PR #309, opened 2024-03-18
[postgres] Remove casting from primary key values query

nathan-artie:
Since the only [two columns type that require casting ](https://github.com/artie-labs/reader/blob/970631c7c8f25b34ad687eec142308e2e739efb7/lib/postgres/scanner.go#L140)(`schema.Array`, `schema.TimeWithTimeZone`) are [not supported as primary key types](https://github.com/artie-labs/reader/blob/970631c7c8f25b34ad687eec142308e2e739efb7/lib/postgres/scanner.go#L19) we can remove the logic that casts columns when we fetch the primary key values.

### DIFF
--- a/lib/postgres/schema/schema.go
+++ b/lib/postgres/schema/schema.go
@@ -198,9 +198,9 @@ type buildPkValuesQueryArgs struct {
 }
 
 func buildPkValuesQuery(args buildPkValuesQueryArgs) string {
-	castedColumns := make([]string, len(args.Keys))
+	escapedColumns := make([]string, len(args.Keys))
 	for i, col := range args.Keys {
-		castedColumns[i] = pgx.Identifier{col.Name}.Sanitize()
+		escapedColumns[i] = pgx.Identifier{col.Name}.Sanitize()
 	}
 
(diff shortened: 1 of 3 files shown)


lib/postgres/schema/schema.go
```diff
@@ -195,17 +195,12 @@ type buildPkValuesQueryArgs struct {
...
-func buildPkValuesQuery(args buildPkValuesQueryArgs) (string, error) {
+func buildPkValuesQuery(args buildPkValuesQueryArgs) string {
 	castedColumns := make([]string, len(args.Keys))
 	for i, col := range args.Keys {
-		var err error
-		castedColumns[i], err = args.CastFunc(col)
-		if err != nil {
-			return "", err
-		}
+		castedColumns[i] = pgx.Identifier{col.Name}.Sanitize()
```
nathan-artie, line 203: From https://github.com/artie-labs/reader/blob/970631c7c8f25b34ad687eec142308e2e739efb7/lib/postgres/scanner.go#L141

Tang8330 (approved):

PR #314, opened 2024-03-22
Fix MySQL Kafka Topic

Tang8330:
MySQL topic didn't include the database name

### DIFF
--- a/integration_tests/mysql/main.go
+++ b/integration_tests/mysql/main.go
@@ -49,22 +49,22 @@ func main() {
 		logger.Fatal("Unable to change sql_mode", slog.Any("err", err))
 	}
 
-	if err = testTypes(db); err != nil {
+	if err = testTypes(db, mysqlCfg.Database); err != nil {
 		logger.Fatal("Types test failed", slog.Any("err", err))
 	}
 
-	if err = testScan(db); err != nil {
+	if err = testScan(db, mysqlCfg.Database); err != nil {
 		logger.Fatal("Scan test failed", slog.Any("err", err))
 	}
 }
 
-func readTable(db *sql.DB, tableName string, batchSize int) ([]lib.RawMessage, error) {
+func readTable(db *sql.DB, dbName, tableName string, batchSize int) ([]lib.RawMessage, error) {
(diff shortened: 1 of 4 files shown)


sources/mysql/snapshot.go
```diff
@@ -50,7 +52,7 @@ func (s Source) snapshotTable(ctx context.Context, writer kafkalib.BatchWriter,
 	logger := slog.With(slog.String("table", tableCfg.Name))
 	snapshotStartTime := time.Now()
 
-	adapter, err := adapter.NewMySQLAdapter(s.db, tableCfg)
+	adapter, err := adapter.NewMySQLAdapter(s.db, s.dbName, tableCfg)
```
nathan-artie, line 55: Can just do `s.cfg.Database` here.

nathan-artie: Oh and maybe we should add it to `slog.String("table", tableCfg.Name)`.

Tang8330: Yep I see it thx

Tang8330: Sure

integration_tests/mysql/main.go
```diff
@@ -49,22 +49,22 @@ func main() {
...
 
-func readTable(db *sql.DB, tableName string, batchSize int) ([]lib.RawMessage, error) {
+func readTable(db *sql.DB, dbName, tableName string, batchSize int) ([]lib.RawMessage, error) {
 	tableCfg := config.MySQLTable{
 		Name:      tableName,
 		BatchSize: uint(batchSize),
 	}
 
-	dbzAdapter, err := adapter.NewMySQLAdapter(db, tableCfg)
+	dbzAdapter, err := adapter.NewMySQLAdapter(db, dbName, tableCfg)
```
nathan-artie, line 67: Could just hardcode "mysql" here since it isn't actually used for any of the tests here, then we don't have to pass it around.

nathan-artie (approved):

integration_tests/mysql/main.go
Tang8330: !

PR #319, opened 2024-03-28
Pull out iteration writing logic

nathan-artie:


sources/source.go
```diff
@@ -8,5 +8,5 @@ import (
 
 type Source interface {
 	Close() error
-	Run(ctx context.Context, writer destinations.DestinationWriter) error
+	Run(ctx context.Context, destination destinations.Destination) error
```
Tang8330, line 11: Thoughts on passing in writer instead of destination? This way, we only need to instantiate the writer once, as opposed to in every source.

Tang8330 (approved):
lgtm % c

sources/source.go
nathan-artie: That was actually what I had originally, and I'd very much like to, but we can't right now because DynamoDB expose an iterator it calls `WriteRawMessages` directly. I'll take a look at switching it.

PR #321, opened 2024-03-29
Pass `writer.Writer` instead of `destinations.Destination`

nathan-artie:


lib/writer/writer.go
```diff
@@ -37,12 +37,12 @@ func (w *Writer) Write(ctx context.Context, iter RawMessageIterator) (int, error
 				return 0, fmt.Errorf("failed to write messages: %w", err)
 			}
 			count += len(msgs)
+			slog.Info("Write progress",
+				slog.Duration("timing", time.Since(start)),
+				slog.Int("batchSize", len(msgs)),
+				slog.Int("total", count),
+			)
 		}
-		slog.Info("Write progress",
```
nathan-artie, line 41: Otherwise it's too noisy if there are a lot of empty batches.

Tang8330 (approved):

PR #322, opened 2024-03-29
[dynamodb] Add max attempts to `processShard`

nathan-artie:
Due to a [bug with `jitter.Jitter`](https://github.com/artie-labs/transfer/pull/395) when we call it with attempts > 58 it'll panic. I fixed the bug in transfer, but it seems smart to also build in a maximum amount of attempts here so that we don't loop forever.

### DIFF
--- a/sources/dynamodb/shard.go
+++ b/sources/dynamodb/shard.go
@@ -2,6 +2,7 @@ package dynamodb
 
 import (
 	"context"
+	"fmt"
 	"log/slog"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sources/dynamodb/shard.go
```diff
@@ -15,6 +16,8 @@ import (
 	"github.com/artie-labs/reader/lib/writer"
 )
 
+const maxAttempts = 50
```
nathan-artie, line 19: @Tang8330 what do you think we should set this to?

Tang8330: Why don't we fix the bug in Jitter instead?

PR #323, opened 2024-03-30
Clean up iterators

nathan-artie:
- Consolidate iterators under `lib/iterator/`
- Add `Iterator[T]` interface


lib/iterator/prebatched.go
```diff
@@ -0,0 +1,31 @@
+package iterator
+
+import "fmt"
+
+type prebatchedIterator[T any] struct {
```
nathan-artie, line 5: Moved from https://github.com/artie-labs/reader/pull/323/files#diff-fae3e6133490758f44967d974676cc3b0213ea320243223e0e1cbe0f842e7d20L44

Tang8330 (approved):

PR #324, opened 2024-03-30
[kafka] Combine `BatchWriter` methods

nathan-artie:


lib/kafkalib/writer.go
```diff
@@ -21,6 +24,36 @@ const (
 	maxJitterMs  = 5000
 )
 
+func newWriter(ctx context.Context, cfg config.Kafka) (*kafka.Writer, error) {
```
nathan-artie, line 27: Just moved this, no changes.

PR #325, opened 2024-04-01
[MySQL] Support Unsigned integers.

Tang8330:
This PR adds support for unsigned integers by making them into `int64`, which is what Debezium is doing.

### DIFF
--- a/integration_tests/mysql/main.go
+++ b/integration_tests/mysql/main.go
@@ -217,7 +217,6 @@ const expectedPayloadTemplate = `{
 						"type": "int64",
 						"optional": false,
 						"default": null,
-						"field": "c_bigint",
 						"field": "c_unsigned_int",
 						"name": "",
 						"parameters": null
--- a/integration_tests/mysql/main.go
+++ b/integration_tests/mysql/main.go
@@ -218,6 +218,15 @@ const expectedPayloadTemplate = `{
 						"optional": false,
 						"default": null,
 						"field": "c_bigint",
+						"field": "c_unsigned_int",
+						"name": "",
+						"parameters": null
+					},
+					{
+						"type": "int64",
+						"optional": false,
+						"default": null,
+						"field": "c_bigint",
 						"name": "",
 						"parameters": null
 					},
@@ -427,6 +436,7 @@ const expectedPayloadTemplate = `{
 			"c_time": 14706000000,
 			"c_timestamp": "2001-02-03T04:05:06Z",
 			"c_tinyint": 1,
+			"c_unsigned_int": 55,
 			"c_varbinary": "Qk5N",
 			"c_varchar": "GHJKL",
 			"c_year": 2001,
@@ -445,8 +455,8 @@ const expectedPayloadTemplate = `{
 
 // testTypes checks that MySQL data types are handled correctly.
 func testTypes(db *sql.DB, dbName string) error {
-	tempTableName, _ := utils.CreateTemporaryTable(db, testTypesCreateTableQuery)
-	//defer dropTableFunc()
+	tempTableName, dropTableFunc := utils.CreateTemporaryTable(db, testTypesCreateTableQuery)
+	defer dropTableFunc()
 
 	// Check reading an empty table
 	_, err := readTable(db, dbName, tempTableName, 100)
--- a/integration_tests/mysql/main.go
+++ b/integration_tests/mysql/main.go
@@ -76,8 +76,11 @@ const testTypesCreateTableQuery = `
 CREATE TABLE %s (
 	pk INTEGER PRIMARY KEY NOT NULL,
 	c_tinyint TINYINT,
+	c_tinyint_unsigned TINYINT UNSIGNED,
 	c_smallint SMALLINT,
+	c_smallint_unsigned SMALLINT UNSIGNED,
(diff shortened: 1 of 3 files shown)


lib/mysql/schema/schema.go
```diff
@@ -111,6 +111,13 @@ func parseColumnDataType(s string) (DataType, *Opts, error) {
 	parenIndex := strings.Index(s, "(")
 	if parenIndex != -1 {
 		if s[len(s)-1] != ')' {
+			// Make sure the format looks like int (n) unsigned
```
nathan-artie, line 114: Do we want to convert `tinyint unsigned`, `smallint unsigned`, and `mediumint unsigned`, to `BigInt`? How does Debezium handle those types? We could probably just return the next size up for each one when the column is unsigned, so `tinyint unsigned` -> `SmallInt`, `smallint unsigned` -> `MediumInt`, etc.

I think we at least want to convert `tinyint(1) unsigned` to a `Boolean` rather than a `BigInt`. 

If we move this logic to the beginning of the function with something like:
```
unsigned := false
if strings.HasSuffix(s, " unsigned") {
    s = strings.TrimSuffix(s, " unsigned")
    unsigned = true
}
```
then we can ignore the value of `unsigned` for `tinyint(1) unsigned` and use it in the other case statements.

Tang8330: Hm, let's see

Tang8330:
```json
                    {
                        "type": "int16",
                        "optional": true,
                        "field": "tinyint_unsigned"
                    },
                    {
                        "type": "int32",
                        "optional": true,
                        "field": "smallint_unsigned"
                    },
                    {
                        "type": "int32",
                        "optional": true,
                        "field": "medium_int_unsigned"
                    }
```

Tang8330: Specifying `tinyint(1) unsigned`, the (1) is ignored. so it's not a bool and we won't need to worry about it.

lib/mysql/schema/schema.go
```diff
@@ -124,14 +134,30 @@ func parseColumnDataType(s string) (DataType, *Opts, error) {
...
+
 		return TinyInt, nil, nil
 	case "smallint":
+		if unsigned {
+			return Int, nil, nil
+		}
+
 		return SmallInt, nil, nil
 	case "mediumint":
+		if unsigned {
```
nathan-artie, line 149: Tried to leave a comment here but it either failed or got lost. `mediumint` is actually [3 bytes](https://dev.mysql.com/doc/refman/8.3/en/integer-types.html#:~:text=65535-,MEDIUMINT,3,-%2D8388608) so we can return `MediumInt` for it.

lib/mysql/schema/schema.go
```diff
@@ -124,14 +134,30 @@ func parseColumnDataType(s string) (DataType, *Opts, error) {
...
+
 		return MediumInt, nil, nil
 	case "int":
+		if unsigned {
+			return BigInt, nil, nil
+		}
+
 		return Int, nil, nil
-	case "bigint":
+	case "bigint", "int unsigned":
```
nathan-artie, line 160: Don't need `int unsigned` here anymore.

lib/mysql/schema/schema.go
```diff
@@ -124,14 +134,30 @@ func parseColumnDataType(s string) (DataType, *Opts, error) {
 			return Boolean, nil, nil
 		}
 
+		if unsigned {
+			return SmallInt, nil, nil
+		}
+
 		return TinyInt, nil, nil
 	case "smallint":
+		if unsigned {
```
nathan-artie, line 143: Add a couple tests for these cases?

Tang8330:
Yeah, I'm adding integration tests

lib/mysql/schema/schema.go
Tang8330: yep

integration_tests/mysql/main.go
```diff
@@ -76,9 +76,13 @@ const testTypesCreateTableQuery = `
 CREATE TABLE %s (
 	pk INTEGER PRIMARY KEY NOT NULL,
 	c_tinyint TINYINT,
+	c_tinyint_unsigned TINYINT UNSIGNED,
 	c_smallint SMALLINT,
+	c_smallint_unsigned SMALLINT UNSIGNED,
 	c_mediumint MEDIUMINT,
+	c_mediumint_unsigned MEDIUMINT UNSIGNED,
 	c_int INT,
+	c_unsigned_int INT(15) UNSIGNED,
```
nathan-artie, line 85: `c_int_unsigned` for consistency?

nathan-artie (approved):

lib/mysql/schema/schema_test.go
```diff
@@ -41,6 +41,35 @@ func TestParseColumnDataType(t *testing.T) {
...
+			expectedType: SmallInt,
+			expectedOpts: nil,
+		},
+		{
+			input:        "smallint unsigned",
+			expectedType: Int,
+			expectedOpts: nil,
+		},
+		{
+			input:        "mediumint unsigned",
```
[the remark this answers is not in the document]
Tang8330: Yep. Adding it back so the comment makes sense.

PR #326, opened 2024-04-01
Add `iterator.Collect` function

nathan-artie:
Makes testing iterators a lot easier.

### DIFF
--- a/lib/iterator/batch.go
+++ b/lib/iterator/batch.go
@@ -8,8 +8,8 @@ type batchIterator[T any] struct {
 	step  int
 }
 
-// Returns an iterater that splits a list of items into batches of the given step size.
-func NewBatchIterator[T any](items []T, step int) Iterator[[]T] {
+// Batched returns an iterater that splits a list of items into batches of the given step size.
+func Batched[T any](items []T, step int) Iterator[[]T] {
 	return &batchIterator[T]{
 		items: items,
 		index: 0,
--- a/lib/iterator/batch.go
+++ b/lib/iterator/batch.go
@@ -8,7 +8,7 @@ type batchIterator[T any] struct {
 	step  int
 }
 
-// Batched returns an iterater that splits a list of items into batches of the given step size.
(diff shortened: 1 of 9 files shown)


lib/iterator/types.go
```diff
@@ -1,6 +0,0 @@
-package iterator
```
nathan-artie, line 1: Renamed this file to `iterator.go`

lib/iterator/iterator.go
```diff
@@ -0,0 +1,19 @@
+package iterator
+
+type Iterator[T any] interface {
```
nathan-artie, line 3: This used to be in `types.go`.

lib/iterator/batch.go
```diff
@@ -8,8 +8,8 @@ type batchIterator[T any] struct {
 	step  int
 }
 
-// Returns an iterater that splits a list of items into batches of the given step size.
-func NewBatchIterator[T any](items []T, step int) Iterator[[]T] {
+// Batched returns an iterater that splits a list of items into batches of the given step size.
```
Tang8330, line 11: `iterator`

lib/iterator/iterator.go
```diff
@@ -0,0 +1,19 @@
...
+	HasNext() bool
+	Next() (T, error)
+}
+
+// Collect returns a new slice containing all the items from an [Iterator].
+func Collect[T any](iter Iterator[T]) ([]T, error) {
```
Tang8330, line 9: Is this only for tests?

nathan-artie: For now, I can see it being used in the future though.

Tang8330: If so, can we make that more clear?

Tang8330: Running this in Prod over a large data set can cause OOM

Tang8330: I do like this function for testing as it can reduce boilerplate

Tang8330 (approved):

PR #330, opened 2024-04-03
Add functional iterators

nathan-artie:
Go version 1.23 will likely ship with [built-in iterator support](https://go.dev/wiki/RangefuncExperiment). Version 1.22 has iterators but they are disabled unless you pass in a `GOEXPERIMENT ` flag. Switching from struct-based iterators to functional iterators will bring us more in-line with the proposed iterators so that we can switch to them in Go 1.23, plus they require less boilerplate since you don't have to implement both `Next()` and `HasNext()` methods.

### DIFF
--- a/lib/debezium/transformer/transformer_test.go
+++ b/lib/debezium/transformer/transformer_test.go
@@ -73,7 +73,7 @@ func TestDebeziumTransformer_Iteration(t *testing.T) {
 		// Empty iterator
 		transformer, err := NewDebeziumTransformer(mockAdatper{iter: iterator.ForSlice([][]Row{})})
 		assert.NoError(t, err)
-		items, err := iterator.Collect(transformer)
+		items, err := iterator.Collect(iterator.ToFunctionalIterator(transformer))
 		assert.NoError(t, err)
 		assert.Empty(t, items)
 	}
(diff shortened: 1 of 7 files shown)


lib/debezium/transformer/transformer_test.go
```diff
@@ -73,7 +73,7 @@ func TestDebeziumTransformer_Iteration(t *testing.T) {
 		// Empty iterator
 		transformer, err := NewDebeziumTransformer(mockAdatper{iter: iterator.ForSlice([][]Row{})})
 		assert.NoError(t, err)
-		items, err := iterator.Collect(transformer)
+		items, err := iterator.Collect(iterator.ToFunctionalIterator(transformer))
```
nathan-artie, line 76: `ToFunctionalIterator` is a shim I added so that I can migrate our iterators to the new functional-style one at a time. It will go away once the migration is complete.

lib/iterator/batch.go
```diff
@@ -1,32 +1,17 @@
...
-		items: items,
-		index: 0,
-		step:  max(step, 1),
-	}
-}
-
-func (i *batchIterator[T]) HasNext() bool {
-	return i.index < len(i.items)
-}
+func Batched[T any](items []T, step int) FunctionalIterator[[]T] {
```
nathan-artie, line 4: This is a prime example of the new functional-style. We return a function that when called produces the next item in the iterator. The function returns three values: `value`, `error`, and `ok`. When `ok` is false it means that the iterator is exhausted.

lib/iterator/iterator.go
```diff
@@ -7,14 +7,30 @@ type Iterator[T any] interface {
 
 // Collect returns a new slice containing all the items from an [Iterator].
 // Used for testing, use only with iterators containing a finite amount of items that fit in memory.
-func Collect[T any](iter Iterator[T]) ([]T, error) {
+func Collect[T any](iter FunctionalIterator[T]) ([]T, error) {
```
nathan-artie, line 10: This is an example of how to iterate over functional iterators. In future version of go when range-over-function is no longer experimental we'll be able to rewrite it to be:
```
for item, err := iter:
```

PR #331, opened 2024-04-03
Return `cdc.Event` instead of `any`

Tang8330:
This will return the payload in the Artie Event format which will make it easier for us to do subsequent saves.

### DIFF
--- a/integration_tests/utils/utils.go
+++ b/integration_tests/utils/utils.go
@@ -43,12 +43,12 @@ func ReadTable(db *sql.DB, dbzAdapter transformer.Adapter) ([]lib.RawMessage, er
 	return rows, nil
 }
 
-func GetPayload(message lib.RawMessage) util.SchemaEventPayload {
-	payloadTyped, ok := message.Event().(*util.SchemaEventPayload)
+func GetEvent(message lib.RawMessage) util.SchemaEventPayload {
+	event, ok := message.Event().(*util.SchemaEventPayload)
 	if !ok {
-		panic("payload is not of type *util.SchemaEventPayload")
+		panic("event is not of type *util.SchemaEventPayload")
 	}
-	return *payloadTyped
+	return *event
 }
 
 func CheckDifference(name, expected, actual string) bool {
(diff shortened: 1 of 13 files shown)


lib/types.go
```diff
@@ -18,20 +19,20 @@ func NewRawMessage(topicSuffix string, partitionKey map[string]any, payload util
...
 		TopicSuffix:  topicSuffix,
 		PartitionKey: partitionKey,
-		mongoPayload: payload,
+		mongoPayload: &payload,
 		mongo:        true,
 	}
 }
 
-func (r RawMessage) GetPayload() any {
+func (r RawMessage) GetPayload() cdc.Event {
```
nathan-artie, line 35: Can we just make `payload` a `cdc.Event` to begin with, then we don't have to have `both `payload` and `mongoPayload` as well as `mongo`?

Tang8330: Yes we can, nice!

lib/types.go
```diff
@@ -1,40 +1,33 @@
...
 )
 
 type RawMessage struct {
 	TopicSuffix  string
 	PartitionKey map[string]any
-	payload      util.SchemaEventPayload
-	mongoPayload mongo.SchemaEventPayload
-
-	mongo bool
+	payload      cdc.Event
```
nathan-artie, line 12: Nit: Can we call this `event`?

lib/types.go
```diff
@@ -1,40 +1,33 @@
...
 }
 
 func NewRawMessage(topicSuffix string, partitionKey map[string]any, payload util.SchemaEventPayload) RawMessage {
```
nathan-artie, line 15: Can `NewRawMessage` just accept `cdc.Event` and then we don't need `NewMongoMessage`?

lib/types.go
nathan-artie: Should we add getters on `TopicSuffix` and `PartitionKey` too then?

nathan-artie: Could drop the `Get` and call it `Event` since `Get*` to me feels like there might be underlying logic.

Tang8330: Sure

nathan-artie (approved):

integration_tests/utils/utils.go
```diff
@@ -44,11 +44,11 @@ func ReadTable(db *sql.DB, dbzAdapter transformer.Adapter) ([]lib.RawMessage, er
 }
 
 func GetPayload(message lib.RawMessage) util.SchemaEventPayload {
-	payloadTyped, ok := message.GetPayload().(util.SchemaEventPayload)
+	payloadTyped, ok := message.GetEvent().(*util.SchemaEventPayload)
 	if !ok {
-		panic("payload is not of type util.SchemaEventPayload")
+		panic("payload is not of type *util.SchemaEventPayload")
```
nathan-artie, line 49: Nit: `event is not of type`.

PR #332, opened 2024-04-03
More Retryable Errors

Tang8330:
Timeouts can be transient, we should try to retry and re-establish a connection.

### DIFF
--- a/lib/kafkalib/errors.go
+++ b/lib/kafkalib/errors.go
@@ -3,8 +3,9 @@ package kafkalib
 import (
 	"context"
 	"errors"
-	"github.com/segmentio/kafka-go"
 	"strings"
+
+	"github.com/segmentio/kafka-go"
 )
 
 func isExceedMaxMessageBytesErr(err error) bool {
--- a/lib/kafkalib/errors.go
+++ b/lib/kafkalib/errors.go
@@ -10,5 +10,20 @@ func isExceedMaxMessageBytesErr(err error) bool {
 // isRetryableError - returns true if the error is retryable
 // If it's retryable, you need to reload the Kafka client.
 func isRetryableError(err error) bool {
-	return err != nil && strings.Contains(err.Error(), "Topic Authorization Failed: the client is not authorized to access the requested topic")
+	if err == nil {
+		return false
+	}
+
+	retryableErrs := []string{
+		"Topic Authorization Failed: the client is not authorized to access the requested topic",
+		"i/o timeout",
+	}
+
+	for _, retryableErr := range retryableErrs {
+		if strings.Contains(err.Error(), retryableErr) {
+			return true
+		}
+	}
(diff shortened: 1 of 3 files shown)


lib/kafkalib/errors.go
```diff
@@ -10,5 +10,20 @@ func isExceedMaxMessageBytesErr(err error) bool {
 // isRetryableError - returns true if the error is retryable
 // If it's retryable, you need to reload the Kafka client.
 func isRetryableError(err error) bool {
-	return err != nil && strings.Contains(err.Error(), "Topic Authorization Failed: the client is not authorized to access the requested topic")
+	if err == nil {
+		return false
+	}
+
+	retryableErrs := []string{
+		"Topic Authorization Failed: the client is not authorized to access the requested topic",
```
nathan-artie, line 18: Comparing errors using their strings seems brittle to me as the error strings may change next time we upgrade the Kafka library. 

You should be able to compare the error against the actual error type https://github.com/segmentio/kafka-go/blob/ebca72eaee918d303c532feb7ff29afdcd8c2efa/error.go#L44

It also looks like there is a `Temporary` method on the error that could be useful https://github.com/segmentio/kafka-go/blob/ebca72eaee918d303c532feb7ff29afdcd8c2efa/error.go#L138

Tang8330: Yep, I considered that

Tang8330: Leave a TODO for the temporary err?

nathan-artie: The second one is [`DeadlineExceededError`] (https://github.com/golang/go/blob/170862d68377ac04a35223521ce9e450007a6032/src/internal/poll/fd.go#L54)

nathan-artie: Err actually `Exceeded ` from https://pkg.go.dev/context#pkg-variables

nathan-artie (approved):

lib/kafkalib/writer.go
```diff
@@ -4,19 +4,18 @@ import (
...
 	"fmt"
-	"log/slog"
-	"time"
-
-	"github.com/artie-labs/transfer/lib/jitter"
-	"github.com/artie-labs/transfer/lib/size"
 	awsCfg "github.com/aws/aws-sdk-go-v2/config"
-	"github.com/segmentio/kafka-go"
 	"github.com/segmentio/kafka-go/sasl/aws_msk_iam_v2"
+	"log/slog"
```
nathan-artie, line 9: This is in the wrong group.

nathan-artie, line 9: Do we want a line of whitespace between stdlib imports and third-party imports?

Tang8330: Sure

PR #336, opened 2024-04-04
[kafka] Return slice instead of iterator from `Batched`

nathan-artie:
Since we only used `Batched` in one place, and we only ever use it on a finite amount of items, we don't need all the extra complexity of an iterator. Also moved `Batched` to `lib/kafkalib` since that's the only place that uses it.


lib/kafkalib/batch.go
```diff
@@ -0,0 +1,13 @@
+package kafkalib
+
+// Batched splits a slice of items into a slice of step-sized slices.
+func Batched[T any](items []T, step int) [][]T {
```
Tang8330, line 4: Lowercase?

Tang8330 (approved):

lib/kafkalib/batch.go
nathan-artie: Good call.

PR #338, opened 2024-04-04
Combine `lib/writer` with `destinations`

nathan-artie:
- Move both to `writers/`
- Rename `Destination` to `DestinationWriter`
- Add `OnComplete` method to `DestinationWriter`
- Rename `WriteRawMessages` to `Write`


writers/writer_test.go
```diff
@@ -15,14 +15,18 @@ type mockDestination struct {
...
-func (m *mockDestination) WriteRawMessages(ctx context.Context, msgs []lib.RawMessage) error {
+func (m *mockDestination) Write(ctx context.Context, msgs []lib.RawMessage) error {
 	if m.emitError {
 		return fmt.Errorf("test write-raw-messages error")
 	}
 	m.messages = append(m.messages, msgs...)
 	return nil
 }
 
+func (b *mockDestination) OnComplete() error {
```
Tang8330, line 26: m?

nathan-artie: Good catch.

PR #347, opened 2024-04-08
[transfer] Stop JSON marshaling/unmarshaling event payloads

nathan-artie:
Transfer can now handle `int16`/`int32`/`int64` values for integer types, `int64` values and time types, and `[]byte` values for decimal types, so we don't have to marshal and unmarshal each row to JSON anymore.


writers/transfer/writer.go
```diff
@@ -85,11 +50,7 @@ func (w *Writer) Write(_ context.Context, messages []lib.RawMessage) error {
 
 	var events []event.Event
 	for _, message := range messages {
-		evt, err := messageToMemoryEvent(message, w.tc)
-		if err != nil {
-			return err
-		}
-		events = append(events, evt)
+		events = append(events, event.ToMemoryEvent(message.Event(), message.PartitionKey(), w.tc, config.Replication))
```
Tang8330, line 53: Didn't JSON marshal also base64 encode? 

Is that covered too?

nathan-artie: It is for `"org.apache.kafka.connect.data.Decimal"` and `"io.debezium.data.VariableScaleDecimal"`

nathan-artie: Though we probably should handle it for plain `[]byte` values too.

Tang8330: Not sure if I follow - can you elaborate?

nathan-artie: https://github.com/artie-labs/transfer/pull/417

Tang8330: Makes sense

Tang8330 (approved):

PR #354, opened 2024-04-12
[mongodb] Fix issue with payload not being written to transfer

nathan-artie:


sources/mongo/collection.go
```diff
@@ -56,12 +57,12 @@ func (c *collectionScanner) Next() ([]lib.RawMessage, error) {
 			return nil, fmt.Errorf("failed to decode document: %w", err)
 		}
 
-		mgoMsg, err := parseMessage(result)
+		mgoMsg, err := ParseMessage(result)
```
Tang8330, line 60: Why does `ParseMessage` and `ToRawMessage` need to be public?

Tang8330 (approved):
lgtm % c

sources/mongo/collection.go
Tang8330: Oh, I see the tests in `writer_test.go`, sgtm

PR #358, opened 2024-04-15
[DynamoDB] Pass along `before` object if available

Tang8330:
Passing `oldImage` through so we can use it as part of our merge predicate work.

I also rewrote the tests for NewMessage to be more clear.


lib/dynamo/parse_message_test.go
```diff
@@ -1,13 +1,109 @@
...
+						S: ptr.ToString("123"),
+					},
+				},
+				ApproximateCreationDateTime: aws.Time(time.Date(2023, 8, 28, 0, 0, 0, 0, time.UTC)),
+			},
+			EventName: ptr.ToString("REMOVE"),
+		}, "testTable")
+
+		assert.NoError(t, err)
+		assert.Equal(t, map[string]any{"foo": "bar"}, msg.beforeRowData)
```
nathan-artie, line 99: Might check that afterRowData is empty.

nathan-artie (approved):

PR #371, opened 2024-04-29
[MySQL] Handle edge case with datetime

Tang8330:


lib/mysql/schema/convert.go
```diff
@@ -79,6 +79,13 @@ func ConvertValue(value any, colType DataType) (any, error) {
 		if !ok {
 			return nil, fmt.Errorf("expected []byte got %T for value: %v", value, value)
 		}
+
+		if string(bytesValue) == "0000-00-00 00:00:00" {
```
nathan-artie, line 83: Add a test?

nathan-artie (approved):

PR #386, opened 2024-05-16
Support Microsoft SQL Server - The Finale

Tang8330:
The final PR to support Microsoft SQL Server

PR #388, opened 2024-05-16
Supporting MSSQL - Part 1

Tang8330:
## Context
This is going to be a series of PRs to support Microsoft SQL Server as a source. The full PR is here: https://github.com/artie-labs/reader/pull/386

The motivation of splitting this up is to make it easier to review.

## Changes
* Upgrading Artie Transfer
* Adding MSSQL Config
* Adding additional converters

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -65,9 +65,9 @@ type Source string
 const (
 	SourceDynamo     Source = "dynamodb"
 	SourceMongoDB    Source = "mongodb"
+	SourceMSSQL      Source = "mssql"
 	SourceMySQL      Source = "mysql"
 	SourcePostgreSQL Source = "postgresql"
-	SourceMSSQL      Source = "mssql"
 )
 
 type Destination string
--- a/config/config.go
+++ b/config/config.go
@@ -67,6 +67,7 @@ const (
 	SourceMongoDB    Source = "mongodb"
 	SourceMySQL      Source = "mysql"
 	SourcePostgreSQL Source = "postgresql"
+	SourceMSSQL      Source = "mssql"
 )
 
 type Destination string
@@ -82,6 +83,7 @@ type Settings struct {
 	MongoDB    *MongoDB    `yaml:"mongodb,omitempty"`
 	MySQL      *MySQL      `yaml:"mysql,omitempty"`
 	PostgreSQL *PostgreSQL `yaml:"postgresql,omitempty"`
+	MSSQL      *MSSQL      `yaml:"mssql,omitempty"`
 
 	Destination Destination         `yaml:"destination"`
 	Kafka       *Kafka              `yaml:"kafka,omitempty"`
@@ -128,6 +130,14 @@ func (s *Settings) Validate() error {
 
 		if err := s.PostgreSQL.Validate(); err != nil {
 			return fmt.Errorf("postgres validation failed: %w", err)
+		}
+	case SourceMSSQL:
+		if s.MSSQL == nil {
+			return fmt.Errorf("mssql config is nil")
+		}
+
+		if err := s.MSSQL.Validate(); err != nil {
+			return fmt.Errorf("mssql validation failed: %w", err)
 		}
 	default:
 		return fmt.Errorf("invalid source: %q", s.Source)
--- a/config/config.go
+++ b/config/config.go
@@ -115,6 +115,14 @@ func (s *Settings) Validate() error {
 		if err := s.MongoDB.Validate(); err != nil {
 			return fmt.Errorf("mongodb validation failed: %w", err)
 		}
+	case SourceMSSQL:
+		if s.MSSQL == nil {
+			return fmt.Errorf("mssql config is nil")
+		}
+
+		if err := s.MSSQL.Validate(); err != nil {
+			return fmt.Errorf("mssql validation failed: %w", err)
+		}
 	case SourceMySQL:
 		if s.MySQL == nil {
 			return fmt.Errorf("mysql config is nil")
@@ -130,14 +138,6 @@ func (s *Settings) Validate() error {
 
 		if err := s.PostgreSQL.Validate(); err != nil {
 			return fmt.Errorf("postgres validation failed: %w", err)
-		}
-	case SourceMSSQL:
-		if s.MSSQL == nil {
-			return fmt.Errorf("mssql config is nil")
-		}
-
-		if err := s.MSSQL.Validate(); err != nil {
-			return fmt.Errorf("mssql validation failed: %w", err)
 		}
 	default:
 		return fmt.Errorf("invalid source: %q", s.Source)
--- a/config/mssql.go
+++ b/config/mssql.go
@@ -1,0 +1,102 @@
+package config
+
+import (
+	"fmt"
+	"github.com/artie-labs/reader/constants"
+	"github.com/artie-labs/reader/lib/rdbms/scan"
+	"github.com/artie-labs/transfer/lib/stringutil"
+	"math"
+	"net/url"
+	"strings"
+)
+
+type MSSQL struct {
+	Host     string        `yaml:"host"`
+	Port     int           `yaml:"port"`
+	Username string        `yaml:"username"`
+	Password string        `yaml:"password"`
+	Database string        `yaml:"database"`
+	Tables   []*MSSQLTable `yaml:"tables"`
+}
+
+type MSSQLTable struct {
+	Name                       string   `yaml:"name"`
+	Schema                     string   `yaml:"schema"`
+	BatchSize                  uint     `yaml:"batchSize"`
+	OptionalPrimaryKeyValStart string   `yaml:"optionalPrimaryKeyValStart"`
+	OptionalPrimaryKeyValEnd   string   `yaml:"optionalPrimaryKeyValEnd"`
+	ExcludeColumns             []string `yaml:"excludeColumns"`
+}
+
+func (m *MSSQL) ToDSN() string {
+	query := url.Values{}
+	query.Add("database", m.Database)
+
+	u := &url.URL{
+		Scheme:   "sqlserver",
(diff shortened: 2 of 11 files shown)


lib/postgres/scanner.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/artie-labs/reader/lib/rdbms/scan"
 )
 
-var supportedPrimaryKeyDataType []schema.DataType = []schema.DataType{
```
Tang8330, line 19: redundant, so removed

config/config.go
```diff
@@ -129,6 +131,14 @@ func (s *Settings) Validate() error {
 		if err := s.PostgreSQL.Validate(); err != nil {
 			return fmt.Errorf("postgres validation failed: %w", err)
 		}
+	case SourceMSSQL:
```
nathan-artie, line 134: IIRC these are in alphabetical order so we should put this before `SourcePostgreSQL`.

Tang8330: Sure

config/mssql_test.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		assert.Equal(t, uint(1), m.GetBatchSize())
+	}
+}
+
+func TestMSSQLTable_GetOptionalPrimaryKeyValStart(t *testing.T) {
+	{
+		// not set
+		m := &MSSQLTable{}
+		assert.Len(t, m.GetOptionalPrimaryKeyValStart(), 0)
```
nathan-artie, line 158: Could do `assert.Empty`

config/mssql_test.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		assert.Equal(t, []string{"a", "b", "c"}, m.GetOptionalPrimaryKeyValStart())
+	}
+}
+
+func TestMSSQLTable_GetOptionalPrimaryKeyValEnd(t *testing.T) {
+	{
+		// not set
+		m := &MSSQLTable{}
+		assert.Len(t, m.GetOptionalPrimaryKeyValEnd(), 0)
```
nathan-artie, line 173: `assert.Empty`

config/mssql.go
```diff
@@ -0,0 +1,102 @@
...
+type MSSQLTable struct {
+	Name                       string   `yaml:"name"`
+	Schema                     string   `yaml:"schema"`
+	BatchSize                  uint     `yaml:"batchSize"`
+	OptionalPrimaryKeyValStart string   `yaml:"optionalPrimaryKeyValStart"`
+	OptionalPrimaryKeyValEnd   string   `yaml:"optionalPrimaryKeyValEnd"`
+	ExcludeColumns             []string `yaml:"excludeColumns"`
+}
+
+func (m *MSSQL) ToDSN() string {
```
nathan-artie, line 31: Add a test for this method?

lib/debezium/converters/time.go
```diff
@@ -20,13 +42,18 @@ func (MicroTimeConverter) ToField(name string) debezium.Field {
...
 
 func (MicroTimeConverter) Convert(value any) (any, error) {
-	strValue, ok := value.(string)
-	if !ok {
-		return nil, fmt.Errorf("expected string got %T with value: %v", value, value)
-	}
-	timeValue, err := time.Parse(time.TimeOnly, strValue)
-	if err != nil {
-		return nil, err
+	var timeValue time.Time
```
nathan-artie, line 45: Nice.

lib/debezium/converters/time.go
```diff
@@ -8,6 +8,28 @@ import (
 	"github.com/artie-labs/transfer/lib/debezium"
 )
 
+type TimeConverter struct{}
+
+func (TimeConverter) ToField(name string) debezium.Field {
+	return debezium.Field{
```
nathan-artie, line 14: Mind adding:
> Represents the number of milliseconds past midnight, and does not include timezone information.
Here?

Tang8330: It's in Transfer already: https://github.com/artie-labs/transfer/blob/9aed2a27fe5ce89869c8656249fdc87048699427/lib/debezium/types.go#L191

Want me to add it still?

lib/debezium/converters/time.go
```diff
@@ -8,6 +8,28 @@ import (
...
+		FieldName:    name,
+		Type:         debezium.Int32,
+		DebeziumType: debezium.Time,
+	}
+}
+
+func (TimeConverter) Convert(value any) (any, error) {
+	switch timeValue := value.(type) {
+	case time.Time:
+		hours := time.Duration(timeValue.Hour()) * time.Hour
```
nathan-artie, line 24: We could have a tiny helper function in this file:
```
func totalNanoseconds(timeValue time.Time) {
    hours := time.Duration(timeValue.Hour()) * time.Hour
    minutes := time.Duration(timeValue.Minute()) * time.Minute
    seconds := time.Duration(timeValue.Second()) * time.Second
    return (hours + minutes + seconds)
}
```
And then use it in both places we do this type of conversion. Also I think we should be adding in `timeValue.Nanosecond()`.

Tang8330: I'm down. I'll add it.

Tang8330: Added

config/config.go
```diff
@@ -82,6 +83,7 @@ type Settings struct {
 	MongoDB    *MongoDB    `yaml:"mongodb,omitempty"`
 	MySQL      *MySQL      `yaml:"mysql,omitempty"`
 	PostgreSQL *PostgreSQL `yaml:"postgresql,omitempty"`
+	MSSQL      *MSSQL      `yaml:"mssql,omitempty"`
```
nathan-artie, line 86: Alphabetize here too.

lib/debezium/converters/time.go
nathan-artie: No need now, I think the `getTimeDuration(timeValue, time.Millisecond)` makes it obvious.

lib/debezium/converters/time.go
```diff
@@ -8,6 +8,25 @@ import (
...
+		FieldName:    name,
+		Type:         debezium.Int32,
+		DebeziumType: debezium.Time,
+	}
+}
+
+func (TimeConverter) Convert(value any) (any, error) {
+	switch timeValue := value.(type) {
+	case time.Time:
+		return getTimeDuration(timeValue, time.Millisecond), nil
```
nathan-artie, line 24: Nit: Wrap with `int32` so we're outputting the same type as we've defined here https://github.com/artie-labs/reader/pull/388/files#diff-5de60c74c1cf57a3f6e6d17dc239b2e77d2e9b9ac954257b55972fdc22926f0cR16

lib/debezium/converters/time.go
```diff
@@ -86,6 +126,51 @@ func (MicroTimestampConverter) Convert(value any) (any, error) {
...
+
+func (NanoTimestampConverter) Convert(value any) (any, error) {
+	timeValue, ok := value.(time.Time)
+	if !ok {
+		return nil, fmt.Errorf("expected time.Time got %T with value: %v", value, value)
+	}
+	return timeValue.UnixMicro() * 1_000, nil
+}
+
+type TimestampConverter struct{}
```
nathan-artie, line 147: Nit: Feels like we should put this before `MicroTimestampConverter` so that precision increases downward in the file.

lib/debezium/converters/time.go
```diff
@@ -86,6 +126,51 @@ func (MicroTimestampConverter) Convert(value any) (any, error) {
...
+}
+
+func (TimestampConverter) Convert(value any) (any, error) {
+	timeValue, ok := value.(time.Time)
+	if !ok {
+		return nil, fmt.Errorf("expected time.Time got %T with value: %v", value, value)
+	}
+
+	if timeValue.Year() > 9999 || timeValue.Year() < 0 {
+		// Avoid copying this column over because it'll cause a JSON Marshal error:
```
nathan-artie, line 165: I wonder if this is actually true anymore since we're emitting an int64 and not a `time.Time` that is then being JSON marshaled. If Debezium would also emit these values then we could also emit them here for parity.

lib/debezium/converters/time_duration.go
```diff
@@ -0,0 +1,16 @@
...
+func getTimeDuration(timeVal time.Time, timeDuration time.Duration) any {
+	hours := time.Duration(timeVal.Hour()) * time.Hour
+	minutes := time.Duration(timeVal.Minute()) * time.Minute
+	seconds := time.Duration(timeVal.Second()) * time.Second
+
+	switch timeDuration {
+	case time.Millisecond:
+		return int32((hours + minutes + seconds) / timeDuration)
+	default:
+		return int64((hours + minutes + seconds) / timeDuration)
```
nathan-artie, line 14: Could just always emit an `int64` and then cast it to an `int32` in the one place that needs to return an `int32`.

lib/debezium/converters/time_duration.go
```diff
@@ -0,0 +1,16 @@
+package converters
+
+import "time"
+
+func getTimeDuration(timeVal time.Time, timeDuration time.Duration) any {
+	hours := time.Duration(timeVal.Hour()) * time.Hour
+	minutes := time.Duration(timeVal.Minute()) * time.Minute
+	seconds := time.Duration(timeVal.Second()) * time.Second
```
nathan-artie, line 8: Should we add in the nanoseconds as well which would allow us to preserve fractional seconds?

lib/debezium/converters/time.go
nathan-artie: We're also not doing this check for the `MicroTimestampConverter ` or `NanoTimestampConverter`.

lib/debezium/converters/time_duration.go
Tang8330: Sure, I'll add ms, microsec and ns

nathan-artie (approved):

PR #396, opened 2024-05-21
[Kafka] Update error assertion for MessageTooLargeError

Tang8330:
Using `errors.Is` is not capturing the MessageTooLarge error.

This is because the Kafka writer calls: https://github.com/segmentio/kafka-go/blob/ebca72eaee918d303c532feb7ff29afdcd8c2efa/writer.go#L634 which is: https://github.com/segmentio/kafka-go/blob/ebca72eaee918d303c532feb7ff29afdcd8c2efa/error.go#L644-L652

This returns the struct https://github.com/segmentio/kafka-go/blob/ebca72eaee918d303c532feb7ff29afdcd8c2efa/error.go#L644-L652

The tests in the SDK is also using `errors.As`: https://github.com/segmentio/kafka-go/blob/ebca72eaee918d303c532feb7ff29afdcd8c2efa/writer_test.go#L383-L385

### DIFF
--- a/lib/kafkalib/writer.go
+++ b/lib/kafkalib/writer.go
@@ -134,7 +134,6 @@ func (b *BatchWriter) Write(ctx context.Context, rawMsgs []lib.RawMessage) error
 				break
 			}
 
-			fmt.Println("kafkaErr: ", kafkaErr, isExceedMaxMessageBytesErr(kafkaErr))
 			if isExceedMaxMessageBytesErr(kafkaErr) {
 				slog.Info("Skipping this batch since the message size exceeded the server max")
 				kafkaErr = nil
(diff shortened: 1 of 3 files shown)


lib/kafkalib/writer.go
```diff
@@ -134,6 +134,7 @@ func (b *BatchWriter) Write(ctx context.Context, rawMsgs []lib.RawMessage) error
 				break
 			}
 
+			fmt.Println("kafkaErr: ", kafkaErr, isExceedMaxMessageBytesErr(kafkaErr))
```
nathan-artie, line 137: Extra print statement.

nathan-artie (approved):

PR #398, opened 2024-05-24
[MSSQL] To support date, time, datetime primary keys

Tang8330:
MSSQL now supports all data types as primary keys.

I also modified the signature of `BuildQuery` to return an error as the MySQL SDK requires us to explicitly parse out `time.Time`

### DIFF
--- a/lib/mssql/scanner.go
+++ b/lib/mssql/scanner.go
@@ -1,192 +1,192 @@
 package mssql
 
 import (
-    "database/sql"
-    "fmt"
-    "slices"
-    "strings"
-    "time"
+	"database/sql"
+	"fmt"
+	"slices"
+	"strings"
+	"time"
 
-    "github.com/artie-labs/transfer/clients/mssql/dialect"
+	"github.com/artie-labs/transfer/clients/mssql/dialect"
 
-    "github.com/artie-labs/reader/lib/mssql/parse"
-    "github.com/artie-labs/reader/lib/mssql/schema"
-    "github.com/artie-labs/reader/lib/rdbms"
-    "github.com/artie-labs/reader/lib/rdbms/column"
-    "github.com/artie-labs/reader/lib/rdbms/primary_key"
-    "github.com/artie-labs/reader/lib/rdbms/scan"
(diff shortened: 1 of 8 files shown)


lib/mssql/scanner.go
```diff
@@ -63,27 +47,89 @@ type scanAdapter struct {
...
 
-	_column := s.columns[columnIdx]
-	if !slices.Contains(supportedPrimaryKeyDataType, _column.Type) {
-		return nil, fmt.Errorf("DataType(%d) for column %q is not supported for use as a primary key", _column.Type, _column.Name)
-	}
-
-	switch _column.Type {
-	case schema.Bit:
-		return value == "1", nil
+	switch _columnType := s.columns[columnIdx].Type; _columnType {
```
nathan-artie, line 63: Is the `; _columnType` at the end needed?

Tang8330: yep :(

lib/mssql/scanner.go
```diff
@@ -63,27 +47,89 @@ type scanAdapter struct {
...
-func (s scanAdapter) ParsePrimaryKeyValueForOverrides(columnName string, value string) (any, error) {
-	// TODO: Implement Date, Time, Datetime for primary key types.
+func (s scanAdapter) ParsePrimaryKeyValueForOverrides(_ string, value string) (any, error) {
+	// We don't need to cast it at all.
+	return value, nil
+}
+
+// parsePrimaryKeyValues - parse primary key values based on the column type.
+// This is needed because the MSSQL SDK does not support parsing `time.Time`, so we need to do it ourselves.
+func (s scanAdapter) parsePrimaryKeyValues(columnName string, value any) (any, error) {
```
nathan-artie, line 57: Nit: I feel like this could be called `encodePrimaryKeyValue` or `serializePrimaryKeyValue` since we're producing a single value that can be directly used by the driver in a query, parse makes me think of going from a string to a higher level object not the other way around.

nathan-artie (approved):

lib/mssql/scanner.go
Tang8330: Sure

PR #405, opened 2024-06-17
Upgrade Artie Transfer.

Tang8330:


lib/debezium/converters/decimal_test.go
```diff
@@ -1,133 +1,133 @@
 package converters
 
 import (
-	"fmt"
-	"testing"
+    "fmt"
```
Tang8330, line 4: Whitespaces are from me running `go fmt`

nathan-artie (approved):

PR #411, opened 2024-06-22
[MySQL] Support Geometry

Tang8330:
## Change

* Adding support for `Geometry`
* Fix byte order bug with Point

### DIFF
--- a/lib/mysql/schema/convert.go
+++ b/lib/mysql/schema/convert.go
@@ -1,162 +1,179 @@
 package schema
 
 import (
-	"encoding/binary"
-	"fmt"
-	"math"
-	"time"
+    "encoding/binary"
+    "fmt"
+    "math"
+    "time"
 )
 
 const DateTimeFormat = "2006-01-02 15:04:05.999999999"
 
 // ConvertValue takes a value returned from the MySQL driver and converts it to a native Go type.
 func ConvertValue(value any, colType DataType) (any, error) {
-	if value == nil {
-		return nil, nil
-	}
+    if value == nil {
+        return nil, nil
+    }
 
-	switch colType {
-	case Bit:
-		// Bits
-		castValue, ok := value.([]byte)
-		if !ok {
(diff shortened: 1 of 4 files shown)


lib/mysql/schema/convert.go
```diff
@@ -126,8 +126,7 @@ func ConvertValue(value any, colType DataType) (any, error) {
 		}
 
 		// The first four bytes are the SRID.
-
-		if byteOrder := bytes[5]; byteOrder != 1 {
+		if byteOrder := bytes[4]; byteOrder != 1 {
```
Tang8330, line 129: This was wrong, didn't account for the zero index. https://go.dev/play/p/LzrwJJScjVE

Tang8330: @nathan-artie FYI

lib/mysql/schema/convert.go
```diff
@@ -139,6 +138,31 @@ func ConvertValue(value any, colType DataType) (any, error) {
...
+		}
+
+		if len(bytes) != 25 {
+			return nil, fmt.Errorf("expected []byte with length 25, length is %d", len(bytes))
+		}
+
+		var byteOrder binary.ByteOrder
+		switch bytes[4] {
+		case 0:
+			byteOrder = binary.BigEndian
```
nathan-artie, line 154: Is this ever big-endian?

lib/mysql/schema/convert.go
nathan-artie: Good catch!

nathan-artie (approved):

lib/mysql/schema/convert.go
nathan-artie: I don't think that matters though as MySQL always uses little-endian to store wkbs: https://dev.mysql.com/doc/refman/8.4/en/gis-data-formats.html#gis-wkb-format:~:text=The%20byte%2Dorder%20indicator%20byte%20is%201%20because%20MySQL%20stores%20geometries%20as%20little%2Dendian%20values.

PR #412, opened 2024-06-24
[MySQL] Support Linestring

Tang8330:
It's basically a freebie since Debezium treats this as a Geography object as well :)

### DIFF
--- a/lib/mysql/schema/convert.go
+++ b/lib/mysql/schema/convert.go
@@ -144,8 +144,8 @@ func ConvertValue(value any, colType DataType) (any, error) {
 			return nil, fmt.Errorf("expected []byte got %T for value: %v", value, value)
 		}
 
-		if len(bytes) != 25 {
-			return nil, fmt.Errorf("expected []byte with length 25, length is %d", len(bytes))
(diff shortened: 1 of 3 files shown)


lib/mysql/schema/convert.go
```diff
@@ -144,8 +144,8 @@ func ConvertValue(value any, colType DataType) (any, error) {
 			return nil, fmt.Errorf("expected []byte got %T for value: %v", value, value)
 		}
 
-		if len(bytes) != 25 {
-			return nil, fmt.Errorf("expected []byte with length 25, length is %d", len(bytes))
+		if len(bytes) < 25 {
```
Tang8330, line 147: LINESTRING can be more than 25 bytes

lib/mysql/schema/schema.go
```diff
@@ -1,229 +1,231 @@
 package schema
 
 import (
```
Tang8330, line 3: For this one, hide whitespaces since my `go fmt` wasn't working correctly (it is now)

danafallon (approved):

PR #415, opened 2024-06-25
[MySQL] Support additional data types

Tang8330:
## Changes

1. Supporting `multipoint`
2. Support `multilinestring`
3. Support `multipolygon`
4. Support `geometrycollection`

With this PR going out, we are now supporting every MySQL data type!


lib/mysql/schema/schema.go
```diff
@@ -220,8 +220,12 @@ func parseColumnDataType(originalS string) (DataType, *Opts, error) {
 	case "point":
 		return Point, nil, nil
 	case
+		"geomcollection",
```
nathan-artie, line 223: Alphabetize?

integration_tests/mysql/main.go
```diff
@@ -183,7 +187,15 @@ INSERT INTO %s VALUES (
 	-- c_linestring
 		ST_GeomFromText('LINESTRING(0 0, 1 1, 2 2)'),
 	-- c_polygon
-		ST_GeomFromText('POLYGON((0 0, 1 1, 1 0, 0 0))')
+		ST_GeomFromText('POLYGON((0 0, 1 1, 1 0, 0 0))'),
+	-- c_multipoint
```
nathan-artie, line 191: Can we set a non-zero SRID for one of these so that we ensure it's being preserved?

nathan-artie (approved):

integration_tests/mysql/main.go
Tang8330: sure

PR #419, opened 2024-06-26
Adding PersistedMap

Tang8330:
Adding a new storage option called PersistedMap which is necessary for storing offsets for different databases when we implement streaming

### DIFF
--- a/lib/persistedmap/persistedmap.go
+++ b/lib/persistedmap/persistedmap.go
@@ -64,12 +64,12 @@ func (p *PersistedMap) flushRoutine() {
 }
 
 func (p *PersistedMap) flush() error {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
 	if !p.shouldSave {
 		return nil
 	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
 
 	file, err := os.Create(p.filePath)
 	if err != nil {
--- a/lib/persistedmap/persistedmap.go
+++ b/lib/persistedmap/persistedmap.go
@@ -1,0 +1,120 @@
+package persistedmap
+
+import (
+	"fmt"
+	"gopkg.in/yaml.v3"
+	"io"
+	"log/slog"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/artie-labs/reader/lib/logger"
+)
+
+type PersistedMap struct {
+	filePath    string
+	shouldSave  bool
+	mu          sync.RWMutex
+	data        map[string]any
+	flushTicker *time.Ticker
+}
+
+func NewPersistedMap(filePath string) *PersistedMap {
+	persistedMap := &PersistedMap{
+		filePath: filePath,
+		data:     make(map[string]any),
+	}
+
+	if err := persistedMap.loadFromFile(); err != nil {
+		slog.Warn("Failed to load persisted map from filepath, starting a new one...", slog.Any("err", err))
+	}
+
+	persistedMap.flushTicker = time.NewTicker(30 * time.Second)
+	go persistedMap.flushRoutine()
+
+	return persistedMap
+}
+
+func (p *PersistedMap) Set(key string, value any) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.data[key] = value
+	p.shouldSave = true
+}
+
+func (p *PersistedMap) Get(key string) (any, bool) {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+
+	value, isOk := p.data[key]
+	return value, isOk
+}
+
+func (p *PersistedMap) flushRoutine() {
+	for {
+		select {
+		case <-p.flushTicker.C:
+			if err := p.flush(); err != nil {
+				logger.Panic("failed to flush data", slog.Any("err", err))
+			}
+		}
+	}
+}
+
+func (p *PersistedMap) flush() error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if !p.shouldSave {
+		return nil
+	}
+
+	file, err := os.Create(p.filePath)
+	if err != nil {
+		return fmt.Errorf("failed to create file: %w", err)
+	}
+
+	yamlBytes, err := yaml.Marshal(p.data)
+	if err != nil {
+		return fmt.Errorf("failed to marshal data: %w", err)
+	}
+
+	if _, err = file.Write(yamlBytes); err != nil {
+		return fmt.Errorf("failed to write to file: %w", err)
+	}
+
+	defer file.Close()
+
+	p.shouldSave = false
+	return nil
+}
+
+func (p *PersistedMap) loadFromFile() error {
+	file, err := os.Open(p.filePath)
+	if err != nil {
+		return err
+	}
+
+	defer file.Close()
+
+	readBytes, err := io.ReadAll(file)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


lib/persistedmap/persistedmap_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v3"
+	"os"
+	"testing"
+)
+
+func TestPersistedMap_LoadFromFile(t *testing.T) {
+	tmpFile, err := os.CreateTemp("", "persistedmap_test")
```
nathan-artie, line 11: If you use `t.TempDir()` you don't have to create the directory or clean up after the test.

lib/persistedmap/persistedmap.go
```diff
@@ -0,0 +1,117 @@
...
+	yamlBytes, err := yaml.Marshal(p.data)
+	if err != nil {
+		return fmt.Errorf("failed to marshal data: %w", err)
+	}
+
+	if _, err = file.Write(yamlBytes); err != nil {
+		return fmt.Errorf("failed to write to file: %w", err)
+	}
+
+	defer file.Close()
```
nathan-artie, line 85: Should this be after the file is created otherwise if writing fails or marshaling fails the file won't get closed.

Tang8330: yep it should

lib/persistedmap/persistedmap.go
```diff
@@ -0,0 +1,117 @@
...
+func (p *PersistedMap) flushRoutine() {
+	for range p.flushTicker.C {
+		if err := p.flush(); err != nil {
+			logger.Panic("Failed to flush data", slog.Any("err", err))
+		}
+	}
+}
+
+func (p *PersistedMap) flush() error {
+	if !p.shouldSave {
```
nathan-artie, line 64: Should this come after the lock? Otherwise if on another thread `p.data[key] = value` executes on a but `p.shouldSave = true` has not yet, we won't flush the change for another 30 seconds.

lib/persistedmap/persistedmap.go
```diff
@@ -0,0 +1,117 @@
...
+func (p *PersistedMap) Set(key string, value any) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.data[key] = value
+	p.shouldSave = true
+}
+
+func (p *PersistedMap) Get(key string) (any, bool) {
+	p.mu.RLock()
```
nathan-artie, line 48: I wonder if we need to lock this for reads.

Tang8330: wdym?

lib/persistedmap/persistedmap.go
Tang8330: True

lib/persistedmap/persistedmap.go
```diff
@@ -0,0 +1,117 @@
...
+	flushTicker *time.Ticker
+}
+
+func NewPersistedMap(filePath string) *PersistedMap {
+	persistedMap := &PersistedMap{
+		filePath: filePath,
+		data:     make(map[string]any),
+	}
+
+	if err := persistedMap.loadFromFile(); err != nil {
```
nathan-artie, line 29: Is this going to error if the file doesn't exist yet? I feel like it's fine to overwrite the file if it doesn't exist but if it does exist and we can't read it we should fail rather than wipe the data.

lib/persistedmap/persistedmap.go
nathan-artie: I need to read up on Go's concurrency model, but if we're just using a lock to prevent two things from writing to the same data at the same time we don't really care if something reads from the data without acquiring the lock first, do we?

Tang8330: It's not limited to writes, concurrent rw on a map will cause a panic

Tang8330: We're using a RWLock here so we can minimize write locking

lib/persistedmap/persistedmap.go
Tang8330: > Is this going to error if the file doesn't exist yet?

Yes it will. It will create a new one.

> I feel like it's fine to overwrite the file if it doesn't exist but if it does exist and we can't read it we should fail rather than wipe the data.

So you want it to be more selective about error checking?

Tang8330: I don't think it's possible to delete a file that you can't read.

lib/persistedmap/persistedmap.go
nathan-artie: We still need to Unlock the mutex before we return.

Tang8330: ah yeah i moved the defer up

lib/persistedmap/persistedmap.go
nathan-artie: The file isn't going to get deleted but the next timer tick will overwrite the file, so in the case that the file were to get corrupted (or if someone were to edit the yaml file incorrectly) we won't be able to inspect what went wrong.

I think that reading from a file that doesn't exist shouldn't return an error from `loadFromFile` at all, and if there is either a filesystem error or an unmarshaling error we should panic or return an error from this function.

lib/persistedmap/persistedmap.go
```diff
@@ -0,0 +1,117 @@
...
+		return fmt.Errorf("failed to read file: %w", err)
+	}
+
+	var data map[string]any
+	if err = yaml.Unmarshal(readBytes, &data); err != nil {
+		return fmt.Errorf("failed to unmarshal data: %w", err)
+	}
+
+	if data == nil {
+		data = make(map[string]any)
```
nathan-artie, line 110: Can we just return in this case? We don't set `data` to an empty map in the case that the file doesn't exist.

lib/persistedmap/persistedmap.go
Tang8330: Sure

lib/persistedmap/persistedmap.go
```diff
@@ -0,0 +1,118 @@
...
+	p.shouldSave = false
+	return nil
+}
+
+func (p *PersistedMap) loadFromFile() error {
+	file, err := os.Open(p.filePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// If the file doesn't exist, create a new map object and return
+			p.data = make(map[string]any)
```
nathan-artie, line 96: Do we either need the lock here or not on lines 114/115?

Tang8330: We shouldn't need a lock at all, I'll remove it.

lib/persistedmap/persistedmap.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func NewPersistedMap(filePath string) *PersistedMap {
+	persistedMap := &PersistedMap{
+		filePath: filePath,
+		data:     make(map[string]any),
+	}
+
+	if err := persistedMap.loadFromFile(); err != nil {
+		logger.Panic("Failed to load persisted map from filepath", err)
```
nathan-artie, line 30: I think this needs to be either `slog.Any("err", err")` or `"err", err`, otherwise the error won't actually show up as `!BADKEY=err`

nathan-artie (approved):

PR #422, opened 2024-06-27
DynamoDB - Respect lineage

Tang8330:
There is an edge case with DynamoDB where if the same record appeared in two different DynamoDB shards (parent -> child), we are processing them in parallel as opposed to sequentially.

This is mainly an issue if a DynamoDB stream hasn't been consumed in a while since shards are records and time based.

Per [SDK docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html#Streams.Processing):

> Because shards have a lineage (parent and children), **an application must always process a parent shard before it processes a child shard**. This helps ensure that the stream records are also processed in the correct order. (If you use the DynamoDB Streams Kinesis Adapter, this is handled for you. Your application processes the shards and stream records in the correct order. It automatically handles new or expired shards, in addition to shards that split while the application is running. For more information, see [Using the DynamoDB Streams Kinesis adapter to process stream records](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.KCLAdapter.html).)

### DIFF
--- a/sources/dynamodb/shard.go
+++ b/sources/dynamodb/shard.go
@@ -28,10 +28,11 @@ func (s *StreamStore) processShard(ctx context.Context, shard *dynamodbstreams.S
 		return
 	}
 
-	if shard.ParentShardId != nil {
-		parentID := *shard.ParentShardId
-		if s.storage.GetShardProcessing(parentID) && !s.storage.GetShardProcessed(parentID) {
-			slog.Info("Parent shard is being processed, let's sleep 3s and retry", slog.String("shardId", *shard.ShardId))
+	if parentID := shard.ParentShardId; parentID != nil {
+		// If the parent shard exists, is it still being processed? If so, let's wait a bit and then retry.
+		// We must process the parent shard first before processing the child shard.
+		if s.storage.GetShardProcessing(*parentID) && !s.storage.GetShardProcessed(*parentID) {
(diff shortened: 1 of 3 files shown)


sources/dynamodb/shard.go
```diff
@@ -23,13 +23,22 @@ func (s *StreamStore) ListenToChannel(ctx context.Context, writer writers.Writer
...
-
 	// Is there another go-routine processing this shard?
 	if s.storage.GetShardProcessing(*shard.ShardId) {
 		return
 	}
 
+	if parentID := shard.ParentShardId; parentID != nil {
+		// If the parent shard exists, is it still being processed? If so, let's wait a bit and then retry.
+		// We must process the parent shard first before processing the child shard.
+		if s.storage.GetShardProcessing(*parentID) && !s.storage.GetShardProcessed(*parentID) {
```
danafallon, line 34: What if the parent hasn't started processing yet? Is that possible?

Tang8330: Hm interesting. I think we need to sort the shards that are returned since there's [no guarantee](https://repost.aws/questions/QU6by8UGM8Q8miP3pzivCZGA/dynamodb-streams-are-shards-ordered-in-the-stream-description#ANiUwyz43KQJeTBZuDbana_A) it's ordered

danafallon (approved):

PR #428, opened 2024-07-02
MySQL datetime guardrail

Tang8330:
If strict mode isn't enabled, MySQL may allow invalid datetime / timestamps.

### DIFF
--- a/lib/mysql/schema/convert.go
+++ b/lib/mysql/schema/convert.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math"
+	"strings"
 	"time"
 )
 
@@ -82,9 +83,8 @@ func ConvertValue(value any, colType DataType) (any, error) {
 			return nil, fmt.Errorf("expected []byte got %T for value: %v", value, value)
 		}
 
-		if string(bytesValue) == "0000-00-00 00:00:00" {
-			// MySQL supports '0000-00-00 00:00:00' for datetime columns.
-			// We are returning `nil` here because this will fail most Time parsers.
+		if strings.HasSuffix(string(bytesValue), "-00-00 00:00:00") {
+			// If MySQL strict mode isn't turned on, it can allow invalid dates like 2020-00-00 00:00:00 or 0000-00-00 00:00:00
 			return nil, nil
 		}
 
--- a/lib/mysql/schema/convert.go
+++ b/lib/mysql/schema/convert.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math"
+	"strconv"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


lib/mysql/schema/convert.go
```diff
@@ -82,13 +83,13 @@ func ConvertValue(value any, colType DataType) (any, error) {
 			return nil, fmt.Errorf("expected []byte got %T for value: %v", value, value)
 		}
 
-		if string(bytesValue) == "0000-00-00 00:00:00" {
-			// MySQL supports '0000-00-00 00:00:00' for datetime columns.
-			// We are returning `nil` here because this will fail most Time parsers.
+		stringValue := string(bytesValue)
+		if strings.HasSuffix(stringValue, "-00-00 00:00:00") {
```
nathan-artie, line 87: Reading the MySQL docs it seems like it's also possible for the date to be non-zero and the month zero, or vice versa, so maybe what we want to do is lop off just the date, split by hyphens, and then check if any of the three segments is only zeros, something like:
```
func hasNonStrictModeDate(d string) bool {
	if len(d) < 10 {
		return false
	}
	parts := strings.Split(d[:10], "-")
	if len(parts) != 3 {
		return false
	}
	for _, part := range parts {
		value, err := strconv.Atoi(part)
		if err != nil {
			return false
		}
		if value == 0 {
			return true
		}
	}
	return false
}
```

Tang8330: > date to be non-zero and the month zero

Oh god.

Tang8330: Yeah makes sense

lib/mysql/schema/convert_test.go
```diff
@@ -318,3 +318,13 @@ func TestConvertValues(t *testing.T) {
 		assert.Equal(t, []any{int64(1234), "hello world", true}, values)
 	}
 }
+
+func TestHasNonStrictModeDate(t *testing.T) {
+	assert.False(t, hasNonStrictModeDate("2021-01-02"))
```
nathan-artie, line 323: I'd add a test for an empty string, a junk string, and a string with just two hyphens, all of which should return false.

lib/mysql/schema/convert.go
nathan-artie: Not to mention whether it's possible to have a zero day/month but a non-zero h/m/s. 😬

lib/mysql/schema/convert.go
```diff
@@ -184,3 +185,27 @@ func ConvertValues(values []any, cols []Column) error {
 	}
 	return nil
 }
+
+// hasNonStrictModeDate - if strict mode is not enabled, we can end up having invalid datetimes
+func hasNonStrictModeDate(d string) bool {
```
nathan-artie, line 190: Lol, I know I suggested this name, but maybe we could call this `hasNonStrictModeInvalidDate` to be a little clearer.

PR #429, opened 2024-07-03
Support MongoDB streaming

Tang8330:
This PR will enable the initial support for MongoDB streaming.

Once this is merged in, we'll add tests, run manual tests before we flip from experimental to fully supported.

### DIFF
--- a/sources/mongo/streaming.go
+++ b/sources/mongo/streaming.go
@@ -1,0 +1,3 @@
+package mongo
+
+type
--- a/sources/mongo/streaming.go
+++ b/sources/mongo/streaming.go
@@ -4,14 +4,15 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
-	"github.com/artie-labs/reader/config"
-	"github.com/artie-labs/reader/constants"
-	"github.com/artie-labs/reader/lib"
-	"github.com/artie-labs/reader/lib/persistedmap"
+	"time"
+
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
-	"time"
+
+	"github.com/artie-labs/reader/config"
+	"github.com/artie-labs/reader/lib"
+	"github.com/artie-labs/reader/lib/persistedmap"
 )
 
 const offsetKey = "offset"
@@ -67,8 +68,7 @@ func newStreamingIterator(ctx context.Context, db *mongo.Database, cfg config.Mo
 	}
 
 	return &streaming{
-		// TODO: Consider making this configurable
-		batchSize:             constants.DefaultBatchSize,
+		batchSize:             cfg.GetStreamingBatchSize(),
 		db:                    db,
 		cfg:                   cfg,
 		changeStream:          cs,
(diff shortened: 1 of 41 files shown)


sources/mongo/streaming.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (s *streaming) Next() ([]lib.RawMessage, error) {
+	var rawMsgs []lib.RawMessage
+	for s.batchSize > int32(len(rawMsgs)) && s.changeStream.TryNext(s.ctx) {
+		var rawChangeEvent bson.M
+		if err := s.changeStream.Decode(&rawChangeEvent); err != nil {
+			return nil, fmt.Errorf("failed to decode change event: %v", err)
+		}
+
+		s.offsets.Set(offsetKey, base64.StdEncoding.EncodeToString(s.changeStream.ResumeToken()))
```
nathan-artie, line 92: Do we want to set this after parsing the change event?

sources/mongo/streaming.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil, fmt.Errorf("failed to parse change event: %w", err)
+		}
+
+		collection, watching := s.collectionsToWatchMap[changeEvent.Collection()]
+		if !watching {
+			continue
+		}
+
+		msg, err := changeEvent.ToMessage()
+		if err != nil {
```
nathan-artie, line 105: If we get an error here which causes streaming to stop are we going to skip over this message when we resume because we'll have set the resume token?

sources/mongo/streaming.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil, fmt.Errorf("failed to get message: %w", err)
+		}
+
+		rawMessage, err := msg.ToRawMessage(collection, s.cfg.Database)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert message to raw message: %w", err)
```
nathan-artie, line 111: Ditto https://github.com/artie-labs/reader/pull/429/files?diff=unified&w=1#r1694272379 here.

sources/mongo/streaming.go
```diff
@@ -0,0 +1,123 @@
...
+
+		rawMsgs = append(rawMsgs, rawMessage)
+	}
+
+	if len(rawMsgs) == 0 {
+		// If there are no messages, let's sleep a bit before we try again
+		time.Sleep(2 * time.Second)
+	}
+
+	return rawMsgs, nil
```
nathan-artie, line 122: A slightly safer approach to setting the resume token would be to:
1. Iterate over all events until we find a an event (n) that is for a collection we are watching
2. Set the offset to the resume token at (n-1) this way if there is an error we can skip these events again
3. Iterate over the event stream till we either fill the buffer or run out of events
4. Set the offset to the current resume token
5. Return the buffer

nathan-artie (approved):

sources/mongo/streaming.go
nathan-artie: Actually steps 1 and 2 can be combined by setting the resume token after every event we observe that is for a collection that we aren't watching. It's just that once we find an event that we want to emit we should stop setting the resume token till we've filled the buffer.

sources/mongo/streaming.go
Tang8330: I don't see why not!

sources/mongo/streaming.go
Tang8330: We should do what we're doing in DynamoDB

PR #431, opened 2024-07-07
[Postgres] Native support for `TIME WITH TIME ZONE`

Tang8330:
As per title. This is following the same spec as Debezium.


sources/postgres/adapter/adapter.go
```diff
@@ -103,7 +103,9 @@ func valueConverterForType(dataType schema.DataType, opts *schema.Opts) (convert
 		return converters.BytesPassthrough{}, nil
 	case schema.Text, schema.UserDefinedText:
 		return converters.StringPassthrough{}, nil
-	case schema.Time, schema.TimeWithTimeZone:
+	case schema.TimeWithTimeZone:
+		return TimeWithTimezoneConverter{}, nil
```
Tang8330, line 107: This was wrong before

danafallon (approved):

PR #432, opened 2024-07-08
[Kafka] Supporting SCRAM-SHA-512

Tang8330:
By doing this, we will now be able to produce Kafka messages to a Kafka cluster w/ SHA 512 enabled. 

This is required when we're rolling our own Kafka clusters

### DIFF
--- a/lib/kafkalib/writer.go
+++ b/lib/kafkalib/writer.go
@@ -10,9 +10,11 @@ import (
 
 	awsCfg "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/segmentio/kafka-go/sasl/aws_msk_iam_v2"
+	"github.com/segmentio/kafka-go/sasl/scram"
 
 	"github.com/artie-labs/reader/config"
 	"github.com/artie-labs/reader/lib"
+	"github.com/artie-labs/reader/lib/logger"
 	"github.com/artie-labs/reader/lib/mtr"
 	"github.com/artie-labs/transfer/lib/jitter"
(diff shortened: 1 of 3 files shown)


lib/kafkalib/writer.go
```diff
@@ -49,6 +53,17 @@ func newWriter(ctx context.Context, cfg config.Kafka) (*kafka.Writer, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to create scram mechanism: %w", err)
+		}
+
+		writer.Transport = &kafka.Transport{
+			DialTimeout: 10 * time.Second,
+			SASL:        mechanism,
+			TLS:         &tls.Config{},
+		}
 	}
```
danafallon, line 67: Should we add a default case that returns an error?

Tang8330: Default is that it's unauthenticated!

Tang8330: Or do you mean adding a case for `None` and add a default for error?

danafallon: Yeah if the mechanism is `None` or if we add a new mechanism and forget to add handling for it here. Is unauthenticated possible/valid?

Tang8330: Yeah unauthenticated is valid, it's the default option

Tang8330: We also run unauth with our Docker setup

danafallon (approved):

PR #434, opened 2024-07-09
[Kafka] Provide the option to disable TLS

Tang8330:
MSK allows us to use TLS without passing in a certificate.

However, Strimzi does not. If you specify TLS, you must provide a certificate.

### DIFF
--- a/lib/kafkalib/writer.go
+++ b/lib/kafkalib/writer.go
@@ -48,22 +48,35 @@ func newWriter(ctx context.Context, cfg config.Kafka) (*kafka.Writer, error) {
 			return nil, fmt.Errorf("failed to load AWS configuration: %w", err)
 		}
 
-		writer.Transport = &kafka.Transport{
+		transport := &kafka.Transport{
 			DialTimeout: 10 * time.Second,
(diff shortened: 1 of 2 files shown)


lib/kafkalib/writer.go
```diff
@@ -48,22 +48,35 @@ func newWriter(ctx context.Context, cfg config.Kafka) (*kafka.Writer, error) {
...
 		}
 
-		writer.Transport = &kafka.Transport{
+		transport := &kafka.Transport{
 			DialTimeout: 10 * time.Second,
 			SASL:        mechanism,
 			TLS:         &tls.Config{},
 		}
+
+		if cfg.DisableTLS {
```
nathan-artie, line 75: If you move this after the case statement you only need one of these.

nathan-artie (approved):

lib/kafkalib/writer.go
Tang8330: I tried that, `writer.Transport` is an interface

Tang8330:
Additional context on why MSK doesn't need it, but Strimzi does: https://cloud-native.slack.com/archives/CMH3Q3SNP/p1719834348714689

PR #439, opened 2024-07-19
Upgrading Artie Transfer

Tang8330:


writers/transfer/writer.go
```diff
@@ -209,6 +209,7 @@ func (w *Writer) flush(reason string) error {
 			tableData.InMemoryColumns().DeleteColumn(constants.DeleteColumnMarker)
 		}
 
+		tableData.InMemoryColumns().DeleteColumn(constants.OnlySetDeleteColumnMarker)
```
Tang8330, line 212: We need this or else append will fail since the number of columns do not match the table

danafallon: Oh good catch, thanks!

danafallon (approved):

PR #441, opened 2024-07-23
[DynamoDB] Pass after schema objects

Tang8330:
We are not passing the after schema object from DynamoDB right now. A side effect of not passing this is that we are not creating primary keys in Dynamo.

This PR adds support of this.

### DIFF
--- a/lib/dynamo/message_test.go
+++ b/lib/dynamo/message_test.go
@@ -1,6 +1,7 @@
 package dynamo
 
 import (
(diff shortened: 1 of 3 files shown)


lib/dynamo/message_test.go
```diff
@@ -1,126 +1,125 @@
 package dynamo
```
Tang8330, line 1: Refactored the test suite to check field type

nathan-artie (approved):

PR #438, opened 2024-07-14
Upgrading to use AWS SDK v2

Tang8330:
As per title. Let's get this merged in after our existing work on DynamoDB and MongoDB streaming.

Things to test manually:

- [x] DynamoDB Streaming
- [x] DynamoDB Snapshot

PR #443, opened 2024-07-23
[DynamoDB] Supporting bytes and bytes set

Tang8330:
While reviewing https://github.com/artie-labs/reader/pull/438, I noticed that we do not support `B` or `BS`. 

As such, this PR adds support for it. I also improved our test script to be easier to run.

### DIFF
--- a/lib/dynamo/message_test.go
+++ b/lib/dynamo/message_test.go
@@ -28,6 +28,26 @@ func TestTransformAttributeValue(t *testing.T) {
 		assert.Equal(t, debezium.Float, fieldType)
 	}
 	{
+		// Bytes
+		actualValue, fieldType, err := transformAttributeValue(&dynamodb.AttributeValue{
+			B: []byte("hello"),
+		})
+		assert.NoError(t, err, err)
+		assert.Equal(t, []byte("hello"), actualValue)
+		assert.Equal(t, debezium.Bytes, fieldType)
+	}
+	{
+		// Bytes set
+		actualValue, fieldType, err := transformAttributeValue(&dynamodb.AttributeValue{
+			BS: [][]byte{
+				[]byte("hello"),
+			},
+		})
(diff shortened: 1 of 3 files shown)


lib/dynamo/message_test.go
```diff
@@ -27,6 +27,27 @@ func TestTransformAttributeValue(t *testing.T) {
 		assert.Equal(t, float64(123), actualValue)
 		assert.Equal(t, debezium.Float, fieldType)
 	}
+	{
+		// Bytes
+		actualValue, fieldType, err := transformAttributeValue(&dynamodb.AttributeValue{
+			B: []byte("hello"),
+		})
+		assert.NoError(t, err, err)
```
nathan-artie, line 35: Don't need the second `err`.

lib/dynamo/message_test.go
```diff
@@ -27,6 +27,27 @@ func TestTransformAttributeValue(t *testing.T) {
...
+	}
+	{
+		// Bytes set
+		actualValue, fieldType, err := transformAttributeValue(&dynamodb.AttributeValue{
+			BS: [][]byte{
+				[]byte("hello"),
+				[]byte("world"),
+			},
+		})
+		assert.NoError(t, err, err)
```
nathan-artie, line 47: Don't need the second `err`.

scripts/dynamo/load.go
```diff
@@ -18,15 +19,16 @@ const (
 	region       = "us-east-1"
 	table        = "ddb-test"
 	maxBatchSize = 25 // DynamoDB's limit for batch write
+	offset       = 500000
```
nathan-artie, line 22: Do we always want offset to be 50k?

Tang8330: I've been changing it each time I run the load test, I can move this to an arg instead

nathan-artie (approved):

PR #446, opened 2024-07-29
[MongoDB] Adding tests

Tang8330:
Adding test coverage to `NewChangeEvent` and `ToMessage`


sources/mongo/change_event.go
```diff
@@ -7,15 +7,13 @@ import (
 
 type ChangeEvent struct {
 	operationType string
-	documentKey   bson.M
```
Tang8330, line 10: We don;'t need it

nathan-artie (approved):

PR #447, opened 2024-07-30
[MongoDB] Improve casting primary key efficiency

Tang8330:
This PR works to minimize the amount of marshalling and unmarshalling we need to do just to generate the JSON extended string for the primary key.

We are doing this by natively supporting and converting the most common data types and falling back to marshaling if we need to

### DIFF
--- a/sources/mongo/message.go
+++ b/sources/mongo/message.go
@@ -3,6 +3,7 @@ package mongo
 import (
 	"encoding/json"
 	"fmt"
+	"go.mongodb.org/mongo-driver/bson/primitive"
 	"time"
 
 	"github.com/artie-labs/reader/config"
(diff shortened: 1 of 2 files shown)


sources/mongo/message.go
```diff
@@ -46,26 +47,43 @@ func ParseMessage(result bson.M, op string) (*Message, error) {
 		return nil, fmt.Errorf("failed to marshal document to JSON extended: %w", err)
 	}
 
-	var jsonExtendedMap map[string]any
-	if err = json.Unmarshal(jsonExtendedBytes, &jsonExtendedMap); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal JSON extended to map: %w", err)
-	}
-
-	pk, isOk := jsonExtendedMap["_id"]
+	bsonPk, isOk := result["_id"]
```
danafallon, line 50: Could do this first so we don't bother marshaling it (line 45) if it's going to error out anyway

danafallon (approved):

PR #455, opened 2024-08-01
[DynamoDB] Avoid OOMing from snapshotting

Tang8330:
Right now, we're reading the entire `json.gz` file and then writing it.

DynamoDB chunks the files for 1.5M rows per file by default. However, if the row is really large - we're susceptible to OOM errors


Tang8330:
Hmm, closing this. I think we need to throttle `StreamJsonGzipFile`, and this doesn't do that exactly.

Tang8330:
Actually, this is fine. We are using an unbuffered channel.

> Unbuffered channels provide synchronous communication, meaning the sender will block until a receiver is ready to receive the value, and vice versa.

danafallon (approved):

PR #458, opened 2024-08-04
[Part 1] Upgrading S3 to use AWS SDK v2

Tang8330:
## Context

We have plans to fully migrate off AWS SDK v1 to v2, however the scope is pretty large and there are breaking differences between v1 and v2 for the DynamoDB export API.

As a result, we're chunking this into separate PRs to make it easier to test and review.

This PR is moving off the v1 SDK for S3.


nathan-artie (approved):

sources/dynamodb/dynamodb.go
```diff
@@ -35,12 +39,21 @@ func Load(cfg config.DynamoDB) (sources.Source, bool, error) {
 	}
 
 	if cfg.Snapshot {
+		_awsCfg, err := awsCfg.LoadDefaultConfig(
+			ctx,
+			awsCfg.WithRegion(cfg.AwsRegion),
```
nathan-artie, line 44: Do we need the region for S3?

Tang8330: Pretty sure we do, but I'll confirm

Tang8330: Yup, we need the region. Otherwise, it will try to load the default region (from your `~/.aws/config` or ENV VAR) and if the bucket is not in that region, it will 301

PR #459, opened 2024-08-05
[MongoDB] Adding an option to disable TLS

Tang8330:
We need the ability to disable TLS so we can support DocumentDB.

### DIFF
--- a/sources/mongo/mongo.go
+++ b/sources/mongo/mongo.go
@@ -26,7 +26,11 @@ func Load(cfg config.MongoDB) (*Source, bool, error) {
 		Password: cfg.Password,
 	}
 
-	opts := options.Client().ApplyURI(cfg.Host).SetAuth(creds).SetTLSConfig(&tls.Config{})
+	opts := options.Client().ApplyURI(cfg.Host).SetAuth(creds)
+	if !cfg.DisableTLS {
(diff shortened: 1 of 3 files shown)


sources/mongo/mongo.go
```diff
@@ -26,7 +26,11 @@ func Load(cfg config.MongoDB) (*Source, bool, error) {
 		Password: cfg.Password,
 	}
 
-	opts := options.Client().ApplyURI(cfg.Host).SetAuth(creds).SetTLSConfig(&tls.Config{})
+	opts := options.Client().ApplyURI(cfg.Host).SetAuth(creds)
+	if !cfg.DisableTLS {
+		opts.SetTLSConfig(&tls.Config{})
```
nathan-artie, line 31: Does this need to be `opts := opts.SetTLSConfig(&tls.Config{})`?

Tang8330: Nope, but we can add it for more clarity

Tang8330:
```go
func (c *ClientOptions) SetTLSConfig(cfg *tls.Config) *ClientOptions {
	c.TLSConfig = cfg
	return c
}
```

nathan-artie (approved):

PR #463, opened 2024-08-09
Parse DynamoDB ARN

Tang8330:
We can parse the AWS region from the ARN, so we don't need them to specify the region.


nathan-artie (approved):
Nice.

PR #470, opened 2024-08-19
Refactoring DDB snapshot

Tang8330:
DynamoDB snapshot now returns an iterator.

* This adds consistency (how all of the sources have been snapshotting)
* No more fixed batches / OOM errors


lib/s3lib/s3lib.go
```diff
@@ -72,10 +68,19 @@ func (s *S3Client) ListFiles(ctx context.Context, fp string) ([]S3File, error) {
 // StreamJsonGzipFile - will take a S3 File that is in `json.gz` format from DynamoDB's export to S3
 // It's not a typical JSON file in that it is compressed and it's new line delimited via separated via an array
 // Which means we can stream this file row by row to not OOM.
-func (s *S3Client) StreamJsonGzipFile(ctx context.Context, file S3File, ch chan<- map[string]types.AttributeValue) error {
-	const maxBufferSize = 1024 * 1024 // 1 MB or adjust as needed
-
+func (s *S3Client) StreamJsonGzipFile(ctx context.Context, files []S3File, ch chan<- map[string]types.AttributeValue) error {
```
Tang8330, line 71: This now takes a list of files

sources/dynamodb/snapshot.go
```diff
@@ -31,15 +28,33 @@ func (s *SnapshotStore) Close() error {
...
 	}
 
-	if err := s.streamAndPublish(ctx, writer); err != nil {
-		return fmt.Errorf("stream and publish failed: %w", err)
+	keys, err := s.retrievePrimaryKeys(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve primary keys: %w", err)
+	}
+
+	ch := make(chan map[string]types.AttributeValue)
```
Tang8330, line 41: Two functionality changes

1. We no longer require running `onComplete` outside of `writer.Write`, this is because:
2. We are using one channel for all the files now, as opposed to one channel per file

lib/s3lib/s3lib.go
danafallon: Should we rename it to `StreamJsonGzipFiles`?

Tang8330: Sounds good :)

sources/dynamodb/snapshot_iterator.go
```diff
@@ -0,0 +1,47 @@
...
+func (s *SnapshotIterator) Next() ([]lib.RawMessage, error) {
+	var msgs []lib.RawMessage
+	for msg := range s.ch {
+		dynamoMsg, err := dynamo.NewMessageFromExport(msg, s.keys, s.tableName)
+		if err != nil {
+			return nil, fmt.Errorf("failed to cast message from DynamoDB, msg: %v, err: %w", msg, err)
+		}
+
+		msgs = append(msgs, dynamoMsg.RawMessage())
+		if s.batchSize > int32(len(msgs)) {
```
danafallon, line 40: Should this be `<=` instead of `>`? Because we want to return if we've hit/exceeded the batch size?

Tang8330: Ah yeah, I was following the other examples https://github.com/artie-labs/reader/blob/815a5165ee6ff971da5e7e8b606534a4a3b885ea/sources/mongo/snapshot.go#L76, but the condition is used to decide whether to stay in the loop or not.

Good catch

danafallon (approved):

PR #471, opened 2024-08-19
[Mongo] Moving files around

Tang8330:
Mainly just moving files around


writers/transfer/writer_test.go
```diff
@@ -1,16 +1,14 @@
 package transfer
 
 import (
-	"testing"
-
+	"github.com/artie-labs/reader/config"
+	"github.com/artie-labs/reader/lib/mongo"
```
danafallon, line 5: Move these imports to the bottom?

danafallon (approved):

PR #476, opened 2024-08-23
Move `encodeDecimalWithScale` from transfer

nathan-artie:
Copy `decimalWithNewExponent` and `encodeDecimalWithScale` from transfer so that they can be removed from transfer.

Since `decimalWithNewExponent` is somewhat opinionated (it truncates extra digits rather rounding or returning an error) it makes more sense to have it reader where it is used so that we can decide which behavior is ideal.

### DIFF
--- a/lib/debezium/converters/decimal.go
+++ b/lib/debezium/converters/decimal.go
@@ -2,10 +2,49 @@ package converters
 
 import (
 	"fmt"
+
 	"github.com/artie-labs/transfer/lib/debezium"
 	"github.com/artie-labs/transfer/lib/typing"
 	"github.com/cockroachdb/apd/v3"
 )
+
+// decimalWithNewExponent takes a [*apd.Decimal] and returns a new [*apd.Decimal] with a the given exponent.
+// If the new exponent is less precise then the extra digits will be truncated.
+func decimalWithNewExponent(decimal *apd.Decimal, newExponent int32) *apd.Decimal {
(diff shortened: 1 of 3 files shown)


lib/debezium/converters/decimal.go
```diff
@@ -2,11 +2,50 @@ package converters
 
 import (
 	"fmt"
+
 	"github.com/artie-labs/transfer/lib/debezium"
 	"github.com/artie-labs/transfer/lib/typing"
 	"github.com/cockroachdb/apd/v3"
 )
 
+// decimalWithNewExponent takes a [*apd.Decimal] and returns a new [*apd.Decimal] with a the given exponent.
```
nathan-artie, line 11: No changes to these two methods.

lib/debezium/converters/decimal_test.go
```diff
@@ -5,10 +5,167 @@ import (
 	"testing"
 
 	"github.com/artie-labs/transfer/lib/debezium"
+	"github.com/artie-labs/transfer/lib/numbers"
 	"github.com/artie-labs/transfer/lib/ptr"
+	"github.com/cockroachdb/apd/v3"
 	"github.com/stretchr/testify/assert"
 )
 
+func TestDecimalWithNewExponent(t *testing.T) {
```
nathan-artie, line 14: No changes to these tests other than changing the case of `encodeDecimalWithScale`.

danafallon (approved):

PR #500, opened 2024-09-26
[Postgres] Adding Column allowlist

Tang8330:
Adding the ability to filter out columns that were not on the include list.

Following a similar pattern to [Debezium](https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-property-column-include-list)

### DIFF
--- a/lib/rdbms/column/column_test.go
+++ b/lib/rdbms/column/column_test.go
@@ -162,20 +162,31 @@ func TestFilterOutExcludedColumns(t *testing.T) {
 func TestFilterForIncludedColumns(t *testing.T) {
 	{
 		// Empty `includeNames`
-		value, err := FilterForIncludedColumns([]mockColumn{{Name: "foo"}}, []string{})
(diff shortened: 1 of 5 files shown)


lib/rdbms/column/column_test.go
```diff
@@ -158,3 +158,35 @@ func TestFilterOutExcludedColumns(t *testing.T) {
 		assert.ErrorContains(t, err, `cannot exclude primary key column "bar"`)
 	}
 }
+
+func TestFilterForIncludedColumns(t *testing.T) {
+	{
+		// Empty `includeNames`
+		value, err := FilterForIncludedColumns([]mockColumn{{Name: "foo"}}, []string{}, []string{})
+		assert.NoError(t, err)
+		assert.Equal(t, value, []mockColumn{{Name: "foo"}})
```
danafallon, line 167: Shouldn't this be empty?

danafallon: Oh nevermind, I see that this is handled as a special case

danafallon (approved):
🎉

PR #503, opened 2024-10-03
[Postgres] Improved handling of `bit`

Tang8330:
Supporting Postgres `bit` better, since you can actually specify the character maximum length via `bit(M)`.

Also, `bit` is an alias of `bit(1)`.


lib/debezium/converters/bit.go
```diff
@@ -6,26 +6,48 @@ import (
...
+		return nil, fmt.Errorf("bit converter failed: invalid char max length")
+	case 1:
+		if stringValue == "0" {
+			return false, nil
+		} else if stringValue == "1" {
+			return true, nil
+		}
+		return nil, fmt.Errorf(`string value %q is not in ["0", "1"]`, value)
+	default:
+		return stringValue, nil
```
nathan-artie, line 51: Should probably ensure that `stringValue` is actually of length `b.charMaxLength`.

integration_tests/postgres/main.go
```diff
@@ -639,6 +657,8 @@ const expectedPayloadTemplate = `{
 			"c_bigint": 9009900990099009000,
 			"c_bigserial": 100000123100000123,
 			"c_bit": true,
+			"c_bit1": true,
+			"c_bit5": "10101",
```
nathan-artie, line 661: What's the actual use case for multiple bits? Is it being used for binary data and we should be outputting a bye array instead?

lib/debezium/converters/bit.go
nathan-artie: For the `case 1:` case too.

Tang8330: Sure

integration_tests/postgres/main.go
Tang8330: Not sure what the actual use cases are. I'll check what Debezium does for this though.

Came across this when I was going through all the data types for `pgvector` and bits(m) was part part of an example.

Tang8330:
```
                    {
                        "type": "boolean",
                        "optional": true,
                        "field": "bit_test"
                    },
                    {
                        "type": "boolean",
                        "optional": true,
                        "field": "bit1_test"
                    },
                    {
                        "type": "bytes",
                        "optional": true,
                        "name": "io.debezium.data.Bits",
                        "version": 1,
                        "parameters": {
                            "length": "5"
                        },
                        "field": "bit5_test"
                    }
```

`bit5_test` is outputted as a byte array, I'll adjust

lib/debezium/converters/bit.go
```diff
@@ -1,31 +1,90 @@
...
+		}
+		return nil, fmt.Errorf(`string value %q is not in ["0", "1"]`, value)
+	default:
+		for _, char := range stringValue {
+			if char != '0' && char != '1' {
+				return nil, fmt.Errorf("invalid binary string %q: contains non-binary characters", stringValue)
+			}
+		}
+
+		intValue, err := strconv.ParseInt(stringValue, 2, 64)
```
nathan-artie, line 66: What happens if stringValue is longer than 64?

Tang8330: It'll overflow. Lemme see if I can find a way to not parse

Tang8330: Using `big.Int` now!

nathan-artie (approved):

PR #513, opened 2024-10-09
Add URI to MongoDB config

nathan-artie:


config/mongodb.go
```diff
@@ -60,8 +63,12 @@ func (m MongoDB) GetStreamingBatchSize() int32 {
 }
 
 func (m MongoDB) Validate() error {
-	if stringutil.Empty(m.Host, m.Database) {
-		return fmt.Errorf("one of the MongoDB settings is empty: host or database")
+	if stringutil.Empty(m.Host, m.URI) {
+		return fmt.Errorf("one of the MongoDB settings is empty: host or URI")
```
danafallon, line 67: Do you want this to make sure that either host or URI is provided? I think currently it will error if they're not both provided.

nathan-artie: Oh yes you're totally right.

danafallon (approved):

PR #514, opened 2024-10-09
Use URI for MongoDB integration test

nathan-artie:


integration_tests/mongo/main.go
```diff
@@ -60,6 +50,8 @@ func main() {
 	if err = testTypes(ctx, db, mongoCfg); err != nil {
 		logger.Fatal("Types test failed", slog.Any("err", err))
 	}
+
+	slog.Info("Test succeeded")
```
nathan-artie, line 54: Added these so we can verify test is actually running.

integration_tests/mongo/main.go
```diff
@@ -60,6 +50,8 @@ func main() {
 	if err = testTypes(ctx, db, mongoCfg); err != nil {
 		logger.Fatal("Types test failed", slog.Any("err", err))
 	}
+
+	slog.Info("Test succeeded 😎")
```
nathan-artie, line 54: Added these so we know that the test is running to completion.

danafallon (approved):

PR #521, opened 2024-10-23
WIP - mysql stream

nathan-artie:


integration_tests/mysql/docker-compose.yaml
```diff
@@ -6,3 +6,4 @@ services:
       MYSQL_ROOT_PASSWORD: mysql
     ports:
       - 3306:3306
+    command: --binlog-row-metadata=FULL
```
nathan-artie, line 9: This makes it so that in the binlog change events we get all the column names. Without this the column names are empty.

sources/mysql/stream.go
```diff
@@ -0,0 +1,341 @@
...
+		syncer: replication.NewBinlogSyncer(syncerConfig),
+	}, nil
+}
+
+func (s *StreamingSource) Close() error {
+	return nil
+}
+
+func (s *StreamingSource) Run(ctx context.Context, writer writers.Writer) error {
+	streamer, err := s.syncer.StartSync(mysql.Position{})
```
nathan-artie, line 49: If we want offsets this is where we would pass them in.

sources/mysql/stream.go
```diff
@@ -0,0 +1,341 @@
...
+			replication.WRITE_ROWS_EVENTv2,
+			replication.UPDATE_ROWS_EVENTv2,
+			replication.DELETE_ROWS_EVENTv2:
+
+			rowsEvent, ok := event.Event.(*replication.RowsEvent)
+			if !ok {
+				return fmt.Errorf("unable to cast event to replication.RowsEvent")
+			}
+
+			if string(rowsEvent.Table.Table) != "foo" {
```
nathan-artie, line 71: I was testing with a local table named foo, this is where we need to filter for the tables we care about.

sources/mysql/stream.go
```diff
@@ -0,0 +1,341 @@
...
+				continue
+			}
+
+			messages, err := convertEventToMessages(event.Header, rowsEvent)
+			if err != nil {
+				slog.Error("Failed to convert event to messages", slog.Any("err", err))
+			} else {
+				for i, message := range messages {
+					slog.Info("messages", slog.Int("index", i), slog.Any("event", message.Event()))
```
nathan-artie, line 80: Emit the messages somewhere.

sources/mysql/stream.go
```diff
@@ -0,0 +1,341 @@
...
+	// Column names are only available if `binlog_row_metadata` is set to `FULL`.
+	// They also only work on versions >= MySQL 8.0.1
+	// See: https://dev.mysql.com/doc/refman/8.4/en/replication-options-binary-log.html#sysvar_binlog_row_metadata
+	columnNames := make([]string, len(event.Table.ColumnName))
+	for i, name := range event.Table.ColumnName {
+		columnNames[i] = string(name)
+	}
+
+	collationMap := event.Table.CollationMap()
+	dataTypes := make([]schema.DataType, len(event.Table.ColumnType))
```
nathan-artie, line 109: This is mapping to the data types that we use for backfilling, we may not want to do this though and build our own list of converters.

sources/mysql/adapter/adapter.go
```diff
@@ -88,7 +88,7 @@ func (m MySQLAdapter) PartitionKeys() []string {
 	return m.table.PrimaryKeys
 }
 
-func valueConverterForType(d schema.DataType, opts *schema.Opts) (converters.ValueConverter, error) {
+func ValueConverterForType(d schema.DataType, opts *schema.Opts) (converters.ValueConverter, error) {
```
nathan-artie, line 91: We probably don't want to share these.

PR #531, opened 2024-11-04
Using Transfer BySize

Tang8330:
By using Transfer's BySize, we are able to be more efficient about how we produce Kafka messages.

Previously, if we had 3 messages (a, b, c) where b is too large, we would fail to publish a and b. With BySize, we'll only skip b.


lib/kafkalib/errors.go
```diff
@@ -5,15 +5,6 @@ import (
 	"github.com/segmentio/kafka-go"
 )
 
-func isExceedMaxMessageBytesErr(err error) bool {
```
Tang8330, line 8: We shouldn't need this anymore because Transfer's BySize will handle this for us

nathan-artie (approved):

PR #535, opened 2024-11-05
Fix MySQL snapshot error

Tang8330:
## Problem

We previously mapped `tinyint(1)` to a boolean, this isn't accurate because `tinyint(1)` can accept values that are beyond 0 and 1.

What's unintuitive about this is that you can create a "boolean" column in MySQL:

```sql
ALTER TABLE foo ADD COLUMN bool_col BOOLEAN;
```

Which just maps to a tinyint(1) under the hood.

Later, you can actually do this and it would be valid.

```sql
UPDATE foo SET bool_col = 127;
```

## Further context

The functionality of mapping 0 and 1 to a boolean column isn't going away. If you define the downstream column as a boolean column, Transfer will automatically convert the values for 0, 1 into false / true.


nathan-artie (approved):

lib/debezium/converters/time.go
```diff
@@ -91,10 +93,21 @@ func (DateConverter) Convert(value any) (any, error) {
 	case time.Time:
 		timeValue = castValue
 	case string:
-		if castValue == "0000-00-00" {
-			// MySQL supports '0000-00-00' for date columns
-			return nil, nil
+		parts := strings.Split(castValue, "-")
```
Tang8330, line 96: FYI @nathan-artie I just added this

PR #538, opened 2024-11-07
Initial MSSQL integration tests

Tang8330:
Adding the initial set of MSSQL integration tests

### DIFF
--- a/integration_tests/mssql/main.go
+++ b/integration_tests/mssql/main.go
@@ -11,6 +11,7 @@ import (
 	"os"
 
 	"github.com/lmittmann/tint"
+	_ "github.com/microsoft/go-mssqldb"
 
 	"github.com/artie-labs/reader/config"
 	"github.com/artie-labs/reader/integration_tests/utils"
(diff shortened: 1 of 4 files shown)


integration_tests/mssql/main.go
```diff
@@ -0,0 +1,218 @@
...
+	return utils.ReadTable(dbzAdapter)
+}
+
+// TODO: Test every data type.
+const testTypesCreateTableQuery = `
+CREATE TABLE %s (
+	pk INTEGER PRIMARY KEY NOT NULL,
+	c_bit BIT,
+	c_bytes VARBINARY,
+	c_int16 SMALLINT,
```
nathan-artie, line 70: It might be clearer to name these after the MS SQL column types (e.g. `c_smallint`, `c_int`, `c_bigint`).

nathan-artie (approved):

integration_tests/mssql/main.go
Tang8330: I thought about that, I did it this way so that it can match up with `schema.DataType`

Tang8330: Technically not necessary though, since we should still test tinyint and smallint even though they both map to `Int16`

PR #544, opened 2024-11-13
Add before backfill setting

nathan-artie:


danafallon (approved):

config/config.go
```diff
@@ -114,6 +122,8 @@ type Settings struct {
 
 	Reporting *Reporting `yaml:"reporting"`
 	Metrics   *Metrics   `yaml:"metrics"`
+
+	BeforeBackfill `yaml:"beforeBackfill,omitempty"`
```
danafallon, line 126: Woah can you leave out the type if it's the same as the field name?

nathan-artie: You can, though it wasn't intentional on my part. I'll make it explicit.

PR #552, opened 2024-11-18
MySQL Streaming

Tang8330:
## Changes

1. Splitting DebeziumTransformer out into `DebeziumTransformer` and `DebeziumIterator`


sources/mysql/streaming/schema_history.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/artie-labs/reader/lib/maputil"
+	"github.com/artie-labs/reader/lib/mysql"
+	"github.com/artie-labs/reader/sources/mysql/adapter"
+	"log/slog"
+	"time"
+)
+
+type SchemaHistoryAdapter struct {
+	cfg           config.MySQL
+	schemaHistory map[string]*maputil.MostRecentMap[adapter.Table]
```
Tang8330, line 15: Explore using only one schema adapter per table at a given time.

This is because binlogs are all ordered.

Tang8330, line 15: We should store the schema history similar to how DBZ does it.

Which is ts, raw_ddl.

PR #559, opened 2024-11-21
Reducing MySQL streaming diff

Tang8330:
Pulling out various conversion functions out into a separate PR to minimize the MySQL diff.

### DIFF
--- a/sources/mysql/streaming/converter.go
+++ b/sources/mysql/streaming/converter.go
@@ -27,3 +27,15 @@ func getTimeFromEvent(evt *replication.BinlogEvent) time.Time {
 	// MySQL binlog only has second precision.
 	return time.Unix(int64(evt.Header.Timestamp), 0)
 }
+
+func zipSlicesToMap(keys []string, values []any) (map[string]any, error) {
+	if len(values) != len(keys) {
+		return nil, fmt.Errorf("keys length (%d) is different from values length (%d)", len(keys), len(values))
+	}
+
+	out := map[string]any{}
+	for i, value := range values {
+		out[keys[i]] = value
+	}
+	return out, nil
+}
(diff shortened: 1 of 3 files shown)


sources/mysql/streaming/converter.go
```diff
@@ -17,3 +18,25 @@ func convertHeaderToOperation(evtType replication.EventType) (string, error) {
...
+	if evt == nil {
+		return time.Time{}
+	}
+
+	// MySQL binlog only has second precision.
+	return time.Unix(int64(evt.Header.Timestamp), 0)
+}
+
+// zipSlicesToMap creates a map from two slices, one of keys and one of values.
+func zipSlicesToMap(keys []string, values []any) (map[string]any, error) {
```
nathan-artie, line 32: Could make this generic.

nathan-artie (approved):

sources/mysql/streaming/converter.go
Tang8330: How would you make the key in the map generic?

Tang8330: Also I don't think we need to make the keys generic, unless you do?

nathan-artie: If it's generic then it can be used any time we want to make a map from a slice of keys and a slice of values, so it's more of a general purpose function than one for this specific case.

```
// zipSlicesToMap creates a map from two slices, one of keys and one of values.
func zipSlicesToMap[K comparable, V any](keys []K, values []V) (map[K]V, error) {
	if len(values) != len(keys) {
		return nil, fmt.Errorf("keys length (%d) is different from values length (%d)", len(keys), len(values))
	}

	out := map[K]V{}
	for i, value := range values {
		out[keys[i]] = value
	}
	return out, nil
}
```

Tang8330: Sweet!

PR #564, opened 2024-12-02
Create table fast follow + enabling MySQL

Tang8330:
This is a continuation of https://github.com/artie-labs/reader/pull/563.

This PR adds the following changes:

* Moves the implementation of `BuildTransformerColumn` out into a shared package
* MySQL snapshot is now calling it

### DIFF
--- a/lib/transfer/columns.go
+++ b/lib/transfer/columns.go
@@ -1,0 +1,38 @@
+package transfer
+
+import (
+	"fmt"
+	"github.com/artie-labs/transfer/lib/typing"
+	"github.com/artie-labs/transfer/lib/typing/columns"
+
+	"github.com/artie-labs/reader/lib/debezium/transformer"
+)
+
+type Adapter interface {
+	FieldConverters() []transformer.FieldConverter
+	PartitionKeys() []string
+}
+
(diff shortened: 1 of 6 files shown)


lib/transfer/columns.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/artie-labs/reader/lib/debezium/transformer"
+)
+
+type Adapter interface {
+	FieldConverters() []transformer.FieldConverter
+	PartitionKeys() []string
+}
+
+func BuildTransferColumns(adapter Adapter) ([]columns.Column, error) {
+	var cols columns.Columns
```
Tang8330, line 17: Moved logic out of `postgres/adapter`

lib/transfer/columns_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (m mockAdapter) FieldConverters() []transformer.FieldConverter {
+	return m.fieldConverters
+}
+
+func (m mockAdapter) PartitionKeys() []string {
+	return m.partitionKeys
+}
+
+func TestBuildTransferColumns(t *testing.T) {
```
Tang8330, line 26: Just moved and added a mock adapter, other than that no changes.

nathan-artie (approved):

PR #569, opened 2024-12-03
[MySQL] Persist DDLs

Tang8330:
## Changes

1. Adding the ability to persist DDLs
2. Processing DDLs and coming up with the concept of Schema + Table adapters

### DIFF
--- a/sources/mysql/streaming/ddl.go
+++ b/sources/mysql/streaming/ddl.go
@@ -10,13 +10,13 @@ import (
 )
 
 type Column struct {
-	Name     string
-	DataType string
+	Name       string
+	DataType   string
+	PrimaryKey bool
 }
 
 type TableAdapter struct {
-	columns     []Column
-	primaryKeys []string
+	columns []Column
 }
 
 type SchemaAdapter struct {
@@ -67,8 +67,9 @@ func (s *SchemaAdapter) applyDDL(result antlr.Event) error {
 		var cols []Column
 		for _, col := range castedResult.GetColumns() {
 			cols = append(cols, Column{
-				Name:     col.Name,
-				DataType: col.DataType,
+				Name:       col.Name,
+				PrimaryKey: col.PrimaryKey,
+				DataType:   col.DataType,
 			})
 		}
 
@@ -95,7 +96,13 @@ func (s *SchemaAdapter) applyDDL(result antlr.Event) error {
 		}
 
 		for _, col := range castedResult.GetColumns() {
-			tblAdapter.primaryKeys = append(tblAdapter.primaryKeys, col.Name)
+			columnIdx := slices.IndexFunc(tblAdapter.columns, func(x Column) bool { return x.Name == col.Name })
+			if columnIdx == -1 {
+				return fmt.Errorf("column not found: %q", col.Name)
+			}
+
+			// Update pk
+			tblAdapter.columns[columnIdx].PrimaryKey = true
 		}
 	case antlr.ModifyColumnEvent:
 		tblAdapter, ok := s.adapters[castedResult.GetTable()]
@@ -110,7 +117,6 @@ func (s *SchemaAdapter) applyDDL(result antlr.Event) error {
 			}
 
 			// TODO: Handle position
-
 			tblAdapter.columns[columnIdx].DataType = col.DataType
 		}
 	case antlr.DropColumnsEvent:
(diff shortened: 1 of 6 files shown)


sources/mysql/streaming/ddl.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return i.schemaAdapter.ApplyDDL(query)
+}
+
+func (s *SchemaAdapter) ApplyDDL(query string) error {
+	results, err := antlr.Parse(query)
+	if err != nil {
+		if antlr.IsParseError(err) {
+			return nil
```
nathan-artie, line 51: We're swallowing the errors here?

Tang8330: This is the error: https://github.com/artie-labs/reader/blob/19f0a9402009c881e1e3cb6053824e1472397e8e/lib/antlr/parse.go#L56

Tang8330: It's not really an error, it's more of the fact that we don't support this kind of DDL.

nathan-artie: If we don't support the DDL shouldn't we fail rather than possibly miss a DDL update?

Tang8330: The way I built the DDL Parser is that we are already parsing all the relevant DDL updates. The irrelevant ones have an error being thrown for now.

I can add a warning to it though.

Tang8330: If we wanted to deliberately fail, we would need to comb through all the possible DDL permutations which is not done right now

sources/mysql/streaming/ddl.go
```diff
@@ -0,0 +1,165 @@
...
+			return fmt.Errorf("table not found: %q", castedResult.GetTable())
+		}
+
+		for _, col := range castedResult.GetColumns() {
+			columnIdx := slices.IndexFunc(tblAdapter.columns, func(x Column) bool { return x.Name == col.Name })
+			if columnIdx == -1 {
+				return fmt.Errorf("column not found: %q", col.Name)
+			}
+
+			tblAdapter.columns = append(tblAdapter.columns[:columnIdx], tblAdapter.columns[columnIdx+1:]...)
```
nathan-artie, line 136: Can use `slices.Delete`.

sources/mysql/streaming/ddl.go
```diff
@@ -0,0 +1,165 @@
...
+		s.adapters[castedResult.GetTable()] = tblAdapter
+	case antlr.AddColumnsEvent:
+		tblAdapter, ok := s.adapters[castedResult.GetTable()]
+		if !ok {
+			return fmt.Errorf("table not found: %q", castedResult.GetTable())
+		}
+
+		for _, col := range castedResult.GetColumns() {
+			// Make sure column does not already exist
+			if slices.IndexFunc(tblAdapter.columns, func(x Column) bool { return x.Name == col.Name }) != -1 {
```
nathan-artie, line 148: `slices.ContainsFunc`

sources/mysql/streaming/ddl.go
Tang8330: Hm, thinking more about this - it might be better to explicitly list out the event types that we don't care about rather than use a fallback like this.

I was thinking this way would be safer for Production since it won't just crash and fail. However, if we did miss a DDL event, it would mask the error and make it harder to troubleshoot

sources/mysql/streaming/ddl_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"testing"
+)
+
+func initializeAdapter(t *testing.T) SchemaAdapter {
+	adapter := SchemaAdapter{adapters: map[string]TableAdapter{}}
+	// Create a table first
+	assert.NoError(t, adapter.ApplyDDL("CREATE TABLE test_table (id INT PRIMARY KEY, name VARCHAR(255), email VARCHAR(255));"))
+
+	// Check the table exists
+	assert.Equal(t, 1, len(adapter.adapters))
```
nathan-artie, line 14: `assert.Len`

nathan-artie (approved):

sources/mysql/streaming/ddl.go
Tang8330: Thanks for the flag here, just redid it to list out all the event types.

I suspect we'll need to do more later, but all is well

PR #579, opened 2024-12-06
[MySQL] SchemaAdapter to call `GetCreateTableDDL`

Tang8330:
Calling `CreateTableDDL` for tables that we're replicating that do not exist in the schema history file.


sources/mysql/streaming/iterator.go
```diff
@@ -118,6 +149,14 @@ func (i *Iterator) Next() ([]lib.RawMessage, error) {
 			}
 
 			switch event.Header.EventType {
+			case
```
Tang8330, line 152: Explicitly skipping events that we don't care about so it doesn't clog our logs.

nathan-artie (approved):

PR #580, opened 2024-12-07
[MySQL] Widen converters

Tang8330:
This PR widens our converters to handle the different data types that are coming from the MySQL streaming library.


lib/debezium/converters/passthrough.go
```diff
@@ -68,7 +67,7 @@ func (Int64Passthrough) Convert(value any) (any, error) {
 	return asInt64(value)
 }
 
-// float32, float64 -> float32
+// float32 -> float32
```
Tang8330, line 70: Comment was wrong

nathan-artie (approved):

PR #581, opened 2024-12-09
Simplifying PersistedMap

Tang8330:
This PR simplifies our implementation of `PersistedMap`. 

Before, we had an async go-routine that would save the contents of the persisted map every 15 seconds if it needed to be saved.

Now we are saving it whenever `Set` gets called, removing the go-routines and the need for mutexes.

### DIFF
--- a/lib/storage/persistedmap/persistedmap_test.go
+++ b/lib/storage/persistedmap/persistedmap_test.go
@@ -22,8 +22,6 @@ func TestPersistedMap_LoadFromFile(t *testing.T) {
 
 	// Load the data from the file into PersistedMap
 	pMap := NewPersistedMap[any](tmpFile.Name())
-	pMap.mu.Lock()
-	defer pMap.mu.Unlock()
 	assert.Equal(t, initialData, pMap.data)
 }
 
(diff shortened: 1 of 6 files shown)


writers/writer.go
```diff
@@ -43,7 +44,9 @@ func (w *Writer) Write(ctx context.Context, iter iterator.Iterator[[]lib.RawMess
 
 			// Is it a streaming iterator? if so, let's commit the offset.
 			if streamingIter, isOk := iter.(iterator.StreamingIterator[[]lib.RawMessage]); isOk {
-				streamingIter.CommitOffset()
+				if err = streamingIter.CommitOffset(); err != nil {
+					logger.Panic("Failed to commit offset", slog.Any("err", err))
```
Tang8330, line 48: This is functionality equivalent to what we were doing before, which is to panic if flushing failed.

lib/storage/persistedmap/persistedmap_test.go
```diff
@@ -55,3 +51,18 @@ func TestPersistedMap_Flush(t *testing.T) {
 	assert.Equal(t, 2, val)
 	assert.True(t, isOk)
 }
+
+func BenchmarkNewPersistedMap(b *testing.B) {
```
Tang8330, line 55: Each set took 0.16ms

nathan-artie (approved):

PR #585, opened 2024-12-10
Add Artie generated columns

Tang8330:
With this change https://github.com/artie-labs/reader/pull/563, we are now creating tables for tables that don't yet exist.

However, we missed adding the additional Artie generated columns such as `__artie_updated_at` which takes consideration from inspecting the `topicConfig`. This PR adds that capability.

### DIFF
--- a/writers/transfer/writer.go
+++ b/writers/transfer/writer.go
@@ -28,6 +28,7 @@ import (
 	"github.com/artie-labs/reader/lib/mtr"
 )
 
+// buildColumns - This will append additional columns based on [kafkalib.TopicConfig]
 func buildColumns(cols []columns.Column, tc kafkalib.TopicConfig) []columns.Column {
 	if tc.IncludeArtieUpdatedAt {
 		cols = append(cols, columns.NewColumn(constants.UpdateColumnMarker, typing.TimestampTZ))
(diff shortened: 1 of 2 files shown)


writers/transfer/writer.go
```diff
@@ -113,14 +131,15 @@ func (w *Writer) messageToEvent(message lib.RawMessage) (event.Event, error) {
...
+func (w *Writer) CreateTable(ctx context.Context, tableName string, cols []columns.Column) error {
 	dwh, ok := w.destination.(destination.DataWarehouse)
 	if !ok {
 		// Don't create the table if it's not a data warehouse.
 		return nil
 	}
 
-	createTableSQL, err := ddl.BuildCreateTableSQL(w.cfg.SharedDestinationSettings.ColumnSettings, dwh.Dialect(), w.getTableID(tableName), false, w.cfg.Mode, columns)
+	// We should include additional columns based in the typing config
+	createTableSQL, err := ddl.BuildCreateTableSQL(w.cfg.SharedDestinationSettings.ColumnSettings, dwh.Dialect(), w.getTableID(tableName), false, w.cfg.Mode, buildColumns(cols, w.tc))
```
nathan-artie, line 142: Should we be using `w.tc.tableName` here? Otherwise we're going to create a table using the source table's name but not the destination table's name.

nathan-artie (approved):

writers/transfer/writer.go
Tang8330: True. Nice flag, this will simplify the signature too.

PR #597, opened 2024-12-14
[MySQL] Handling different flavors of `SHOW CREATE TABLE ...` with regards to PKs

Tang8330:
TIL that there's a discrepancy between `SHOW CREATE TABLE <table_name>` for different MySQL versions.

This PR does two things:

1. Adds validation to ensure all emitted rows must have partition keys set
2. Handle the ability to process primary key DDLs where the primary key is not specified inline with the column definition.

### DIFF
--- a/lib/debezium/transformer/light_transformer.go
+++ b/lib/debezium/transformer/light_transformer.go
@@ -39,7 +39,7 @@ func (l LightDebeziumTransformer) BuildPartitionKey(beforeRow, afterRow Row) (ma
 	}
 
 	row := afterRow
-	if afterRow == nil {
+	if len(afterRow) == 0 {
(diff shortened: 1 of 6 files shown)


lib/debezium/transformer/light_transformer.go
```diff
@@ -39,7 +39,7 @@ func (l LightDebeziumTransformer) BuildPartitionKey(beforeRow, afterRow Row) (ma
 	}
 
 	row := afterRow
-	if afterRow == nil {
+	if len(afterRow) == 0 {
```
Tang8330, line 42: The previous way was wrong

nathan-artie (approved):

PR #603, opened 2024-12-17
Support `alter table change column` events for mysql

danafallon:


lib/antlr/alter_table.go
```diff
@@ -122,7 +128,7 @@ func processAddPrimaryKey(ctx *generated.AlterByAddPrimaryKeyContext) ([]Column,
 	return cols, nil
 }
 
-func processAddOrModifyColumn(ctx generated.IAlterSpecificationContext) (Column, error) {
+func processAddOrModifyColumn(ctx generated.IAlterSpecificationContext, isChangeEvent bool) (Column, error) {
```
Tang8330, line 131: Instead of passing this in, can we infer from the `ctx?

```go
_, isChangeEvent := ctx.(*generated.AlterByChangeColumnContext)
```

lib/antlr/alter_table.go
```diff
@@ -161,6 +169,11 @@ func processAddOrModifyColumn(ctx generated.IAlterSpecificationContext) (Column,
 		}
 	}
 
+	if isChangeEvent {
+		// The first name in the list is the column's previous name and isn't relevant here
+		names = names[1:]
```
Tang8330, line 174: Great!

lib/antlr/alter_table.go
```diff
@@ -181,6 +194,24 @@ func processAddOrModifyColumn(ctx generated.IAlterSpecificationContext) (Column,
 	return col, nil
 }
 
+func processChangeColumn(tableName string, spec *generated.AlterByChangeColumnContext) (ModifyColumnEvent, error) {
+	col, err := processAddOrModifyColumn(spec, true)
+	if err != nil {
+		return ModifyColumnEvent{}, err
+	}
+
+	renameEvent, err := processRenameColumn(tableName, spec.AllUid()[:2])
```
Tang8330, line 203: Want to add a check here to make sure `processChangeColumn` has either 2 or 3 columns? If it doesn't have either 2 or 3, then error

Tang8330 (approved):
LGTM % c

lib/antlr/alter_table.go
danafallon: Ooh yep!

PR #606, opened 2024-12-17
[MySQL] GTID scaffold

Tang8330:


sources/mysql/validate.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := row.Scan(&variableName, &value); err != nil {
+		return "", fmt.Errorf("failed to scan row: %w", err)
+	} else if variableName != name {
+		return "", fmt.Errorf("the variable %q was returned instead of %q", variableName, name)
+	}
+
+	return value, nil
+}
+
+func ValidateMySQL(ctx context.Context, db *sql.DB, validateStreaming bool, validateGTID bool) error {
```
Tang8330, line 27: I made this public so that we can reuse the same validation logic in `dataplane-api` without having to add more boilerplate

nathan-artie (approved):

PR #608, opened 2024-12-18
[MysQL] Moving filter columns logic

Tang8330:
Moving the filter columns logic to `buildParsedColumns` so we can filter out columns at a higher level.

Once this PR is in, we'll have column exclusion supported.

### DIFF
--- a/lib/debezium/transformer/light_transformer.go
+++ b/lib/debezium/transformer/light_transformer.go
@@ -2,8 +2,6 @@ package transformer
 
 import (
 	"fmt"
-	"time"
-
 	"github.com/artie-labs/transfer/lib/cdc/util"
 	"github.com/artie-labs/transfer/lib/debezium"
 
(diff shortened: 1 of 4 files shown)


lib/debezium/transformer/light_transformer.go
```diff
@@ -47,7 +45,7 @@ func (l LightDebeziumTransformer) BuildPartitionKey(beforeRow, afterRow Row) (de
 	return convertPartitionKey(l.valueConverters, l.partitionKeys, row)
 }
 
-func (l LightDebeziumTransformer) BuildEventPayload(source util.Source, beforeRow Row, afterRow Row, op string, ts time.Time) (util.SchemaEventPayload, error) {
+func (l LightDebeziumTransformer) BuildEventPayload(source util.Source, beforeRow Row, afterRow Row, op string) (util.SchemaEventPayload, error) {
```
Tang8330, line 48: Unused arg

danafallon (approved):

PR #639, opened 2025-01-06
[Debezium Transformer] Addressing TODO

Tang8330:
This PR is addressing the TODO that stemmed from here: https://github.com/artie-labs/reader/pull/639/files#diff-9868cb345852ed93694bfe7f0611140f441c8243027d6e117e1e867e1f9706ce

The rest of the PR is fixing our tests


lib/debezium/transformer/transformer.go
```diff
@@ -150,7 +146,7 @@ func convertRow(valueConverters map[string]converters.ValueConverter, row Row) (
 func convertPartitionKey(valueConverters map[string]converters.ValueConverter, partitionKeys []string, row Row) (debezium.PrimaryKeyPayload, error) {
 	payload := make(map[string]any, len(partitionKeys))
 	pkFields := make([]debezium.Field, len(partitionKeys))
-	for _, key := range partitionKeys {
+	for index, key := range partitionKeys {
```
Tang8330, line 149: This was wrong, we would end up with extra entries because we initialized an array with empty elements and then appended new ones

nathan-artie (approved):

PR #655, opened 2025-01-14
[MySQL] Improve composite key scanning

Tang8330:


Tang8330:
Putting this back to draft, I think we can get it to be slightly faster.

Tang8330:
Just tested this.

Master ran:
```
Jan 13 18:42:34.573 INF Write progress totalSize=1000000 totalDuration=36.748030333s batchSize=0 batchDuration=220.83575ms
```

This branch:
```
Jan 13 18:41:49.171 INF Write progress totalSize=1000000 totalDuration=15.478943833s batchSize=0 batchDuration=3.644875ms
```

Tang8330:
As fast as it can get I think. Ran explain analyze and it iterated over the range index vs just index.

lib/mysql/scanner/scanner_test.go
```diff
@@ -238,21 +238,67 @@ func TestScanAdapter_BuildQuery(t *testing.T) {
...
+			query, parameters, err := adapter.BuildQuery(keys, false, 12)
+			assert.NoError(t, err)
+			assert.Equal(t, "SELECT `foo`,`bar` FROM `table` WHERE ((`foo` > ?) OR (`foo` = ? AND `bar` > ?)) AND ((`foo` <= ?) OR (`foo` = ? AND `bar` <= ?)) ORDER BY `foo`,`bar` LIMIT 12", query)
+			assert.Equal(t, []any{"a", "a", "c", "b", "b", "d"}, parameters)
+		}
+		{
+			// inclusive upper and lower bounds
+			query, parameters, err := adapter.BuildQuery(keys, true, 12)
+			assert.NoError(t, err)
+			assert.Equal(t, "SELECT `foo`,`bar` FROM `table` WHERE ((`foo` >= ?) OR (`foo` = ? AND `bar` >= ?)) AND ((`foo` <= ?) OR (`foo` = ? AND `bar` <= ?)) ORDER BY `foo`,`bar` LIMIT 12", query)
```
nathan-artie, line 278: For `(`foo` >= ?) OR (`foo` = ? AND `bar` >= ?)` the second condition is redundant because the first condition `(`foo` >= ?)` will match everything that the second condition `(`foo` = ? AND `bar` >= ?)` will match.

Also if we're querying for foo >= 10 and bar >= 20 isn't this query going to include foo = 10 and bar = 5?

Tang8330: Hm, I think the equality symbol is the one that's causing potential trouble. Let me see if I can simplify it to be.

```sql
SELECT `foo`,`bar` FROM `table` WHERE ((`foo` > 0) OR (`foo` = 0 AND `bar` >= 15)) AND ((`foo` < 0) OR (`foo` = 15 AND `bar` <= 20))
```

Tang8330: Marking into a draft so I can improve this more

Tang8330: Just spent more time thinking through this and yes the first condition may seem redundant, the min/max composite keys are not absolute values.

Consider this example:

```sql
mysql> SELECT * FROM artie_reader_17417 ORDER BY `col1`,`col2`;
+------+------+
| col1 | col2 |
+------+------+
|    1 |    2 |
|    3 |    1 |
|    4 |    1 |
|    5 |    1 |
|    6 |    9 |
|    7 |    3 |
|    8 |    3 |
|    9 |    3 |
|   10 |    3 |
+------+------+
9 rows in set (0.01 sec)
```

Retrieving the min key:
```sql
mysql> SELECT `col1`,`col2` FROM `artie_reader_17417` ORDER BY `col1`,`col2` LIMIT 1;
+------+------+
| col1 | col2 |
+------+------+
|    1 |    2 |
+------+------+
```

Retrieving the max key:
```sql
mysql> SELECT `col1`,`col2` FROM `artie_reader_17417` ORDER BY `col1` DESC,`col2` DESC LIMIT 1;
+------+------+
| col1 | col2 |
+------+------+
|   10 |    3 |
+------+------+
```

* Where starting col1 = 1, ending col1 = 10, starting col2 = 2, ending col2 = 3.
* Note that the max is highly dependent on the first column.

This is the proposed query in the PR:
```sql
mysql> SELECT `col1`,`col2`
    -> FROM `artie_reader_17417`
    -> WHERE ((`col1` >= 1) OR (`col1` = 1 AND `col2` >= 2))
    ->   AND ((`col1` <= 10) OR (`col1` = 10 AND `col2` <= 3))
    -> ORDER BY `col1`,`col2`
    -> LIMIT 2;
+------+------+
| col1 | col2 |
+------+------+
|    1 |    2 |
|    3 |    1 |
```

If we removed the additional conditions, we'll end up with something like this:

```
mysql> SELECT `col1`,`col2`
    -> FROM `artie_reader_17417`
    -> WHERE ((`col1` >= 1 AND `col2` >= 2))
    ->   AND ((`col1` <= 10) OR (`col1` = 10 AND `col2` <= 3))
    -> ORDER BY `col1`,`col2`
    -> LIMIT 2;
+------+------+
| col1 | col2 |
+------+------+
|    1 |    2 |
|    6 |    9 |
+------+------+
```

Note that row (col1, col2) (3, 1) was skipped because col2 was less than starting value for col2.

lib/mysql/scanner/scanner.go
```diff
@@ -125,42 +124,119 @@ func (s scanAdapter) ParsePrimaryKeyValueForOverrides(columnName string, value s
 	}
 }
 
-func (s scanAdapter) BuildQuery(primaryKeys []primary_key.Key, isFirstBatch bool, batchSize uint) (string, []any, error) {
-	colNames := make([]string, len(s.columns))
-	for idx, col := range s.columns {
-		colNames[idx] = schema.QuoteIdentifier(col.Name)
+func (s scanAdapter) buildSinglePrimaryKeyQuery(colNames []string, tableNameQuoted string, quotedKeyName string, startingValue, endingValue any, isFirstBatch bool, batchSize uint) (string, []any, error) {
```
Tang8330, line 127: Split this into a separate function so that it's easier to read

lib/mysql/scanner/scanner_test.go
Tang8330: I also added a lot more tests + scenarios to ensure we're scanning everything uniquely. I think it should be good to be reviewed again

PR #656, opened 2025-01-14
[MSSQL] Streaming scaffold

Tang8330:
Starting the streaming scaffold.


config/mssql.go
```diff
@@ -36,6 +46,9 @@ type MSSQLTable struct {
 func (m *MSSQL) ToDSN() string {
 	query := url.Values{}
 	query.Add("database", m.Database)
+	if m.DisableEncryption {
```
danafallon, line 49: Could we check the env here and only allow disabling encryption if it's dev?

Tang8330: Hm, how would you imagine the check?

danafallon: Like:
```go
if m.DisableEncryption && os.Getenv("ENV") == "dev" {
    query.Add("encrypt", "disable")
}
```
Just to make sure we don't allow disabling encryption in prod.

Do we have an env var for environment (dev/prod/etc) in reader?

Tang8330: Ah I see what you mean. We don't have an env var in Reader. I don't mind adding it though

danafallon: Ah k, yeah I think that would be safer to make sure we don't accidentally send customer data unencrypted.

.envrc
```diff
@@ -0,0 +1,6 @@
+export READER_ENV=dev
```
danafallon, line 1: Will this value definitely not get loaded in prod? Should we leave it blank here so you have to explicitly set it in your .envrc-personal?

Tang8330: > Will this value definitely not get loaded in prod?

This only applies to local dev

Tang8330: And you'll need direnv installed for it to load

danafallon (approved):

